PR #159, opened 2021-04-27
AC-496 react localization

melpicazo:
support Spanish language translations for data messenger


src/components/DataMessenger/DataMessenger.js
```diff
@@ -46,6 +47,79 @@ import { runDrilldown, fetchQueryTips } from '../../js/queryService'
 import 'rc-drawer/assets/index.css'
 import './DataMessenger.scss'
 
+let strings = new LocalizedStrings({
```
Ben-CHATA, line 50: Can we change the word "strings" to something like "lang" or "language" to better represent what we're referring to?

melpicazo: changed to lang

Ben-CHATA (approved):

PR #161, opened 2021-04-30
fix tooltip width

AlexLiu-Chata:


src/components/DataMessenger/DataMessenger.js
```diff
@@ -1299,17 +1299,16 @@ export default class DataMessenger extends React.Component {
       'drawer-content-wrapper'
     )[0]
     if (
-      (chartToolTipElement && dataMessenger) ||
-      this.props.placement !== 'top' ||
+      chartToolTipElement &&
+      dataMessenger &&
+      this.props.placement !== 'top' &&
```
Ben-CHATA, line 1304: this placement prop can only be either top or bottom, never both. changing || to && will nullify the whole condition

Ben-CHATA (approved):

PR #174, opened 2021-05-06
fix bug of auto-checking;add edit icon

AlexLiu-Chata:


src/components/Notifications/DataAlertModal/DataAlertModal.js
```diff
@@ -68,7 +67,6 @@ export default class DataAlertModal extends React.Component {
     onManagementCreateDataAlert: () => {},
     onManagementDeleteDataAlert: () => {},
     title: 'Create Data Alert',
-    titleIcon: undefined,
```
Ben-CHATA, line 71: Do we not need these here?

src/components/Notifications/DataAlertModal/DataAlertModal.js
```diff
@@ -49,7 +49,6 @@ export default class DataAlertModal extends React.Component {
     onManagementCreateDataAlert: PropTypes.func,
     onManagementDeleteDataAlert: PropTypes.func,
     title: PropTypes.string,
-    titleIcon: PropTypes.elementType,
```
Ben-CHATA, line 52: Do we not need these here?

src/components/Notifications/DataAlertModal/DataAlertModal.js
AlexLiu-Chata: > Do we not need these here?

No, we do not need. Because you added the ` titleIcon` prop in `<Modal/>`. We can just assign some value for this prop(in `<DataAlertModal/>`)

Ben-CHATA (approved):

PR #180, opened 2021-05-07
fix bug AC-1004;fix prop-type error

AlexLiu-Chata:


src/components/Dashboard/Dashboard.scss
```diff
@@ -40,6 +40,8 @@
     &.bottom {
       position: absolute;
       bottom: 5px;
+      width: 100px;
```
AlexLiu-Chata, line 43: AC-1004

savithChata (approved):

PR #214, opened 2021-07-21
Clear Feature

Lujaina-Chata:
Requiring assistance on the clear button. Problem: linking the form and visual.


Ben-CHATA (changes requested):
Will need to `git pull origin develop` before any merges

Ben-CHATA:
-

PR #308, opened 2022-03-09
add autoAE prop

AlexLiu-Chata:


Ben-CHATA (changes requested):
What is this prop doing? Can we call it something a little bit more descriptive?

PR #340, opened 2022-03-31
fix(dataalerts): disable user edit Project Alerts feature, delete re-…

alexzhou-chata:
…send button for Project Alerts


src/components/Notifications/DataAlerts/DataAlerts.js
```diff
@@ -346,24 +355,27 @@ export default class DataAlerts extends React.Component {
...
-                                this.props.authentication
-                              )
-                              this.getDataAlerts()
-                            }}
-                          >
-                            <span className="react-autoql-re-initialize-btn-text">
-                              <Icon type="warning-triangle" /> Resend
-                            </span>
-                          </Button>
+                          {notification.type === 'CUSTOM' ? (
```
NikkiMeganMoore, line 358: Instead of 
`{notification.type === 'CUSTOM' ? (...) : (null)}`
you could also do
`{notification.type === 'CUSTOM' && (...)}`

PR #354, opened 2022-04-18
Ac 1766 bring new query topics endpoint into the dm widget and change prop function

alexzhou-chata:


example/src/App.js
```diff
@@ -1354,6 +1317,7 @@ export default class App extends Component {
         activeNotificationData={this.state.activeNotificationContent}
         defaultTab={this.state.defaultTab}
         autoChartAggregations={this.state.autoChartAggregations}
+        hasQueryTopics={true}
```
NikkiMeganMoore, line 1320: Since this will be exposed to the public, we may want to stay consistent with the previous naming. Something like:

enableQueryQuickStartTopics

We also need to make sure to update the public docs when this is released. https://chata.readme.io/docs/react-data-messenger#props

src/components/DataMessenger/DataMessenger.js
```diff
@@ -79,6 +80,8 @@ export default class DataMessenger extends React.Component {
       selectedValueLabel: undefined,
       isSizeMaximum: false,
       conditions: undefined,
+      isLoadedTopics: false,
```
NikkiMeganMoore, line 83: hasLoadedTopics

src/components/DataMessenger/DataMessenger.js
```diff
@@ -193,15 +196,29 @@ export default class DataMessenger extends React.Component {
       console.error(error)
       this.setState({ hasError: true })
     }
+    if (this.props.hasQueryTopics) {
+      fetchConditions(getAuthentication(this.props.authentication))
```
NikkiMeganMoore, line 200: Why do you need to check for "hasQueryTopics" to fetch filter locks?

src/components/DataMessenger/DataMessenger.js
```diff
@@ -241,7 +258,12 @@ export default class DataMessenger extends React.Component {
         this.clearMessages()
       }
       if (this.state.messages.length === 0) {
-        this.setintroMessages()
+        const { isLoadedTopics } = this.state
+        if (isLoadedTopics) {
+          this.setintroMessages()
```
NikkiMeganMoore, line 263: Both of these blocks have the same this.setintroMessages(), why have an if statement at all?

Also can we change this to setIntroMessages with the capital "i"   must have been a typo the first time

src/components/DataMessenger/DataMessenger.js
```diff
@@ -421,9 +442,8 @@ export default class DataMessenger extends React.Component {
       }),
     ]
 
-    if (_get(this.props.queryQuickStartTopics, 'length')) {
+    if (this.state.topics.length > 0) {
```
NikkiMeganMoore, line 445: May want to turn this into _get(this.state, 'topics.length') just in case something goes wrong with the api call?

NikkiMeganMoore (changes requested):

src/components/DataMessenger/DataMessenger.js
```diff
@@ -207,6 +210,26 @@ export default class DataMessenger extends React.Component {
...
         console.error(error)
       })
+
+    if (this.props.enableQueryQuickStartTopics) {
+      fetchTopics(getAuthentication(this.props.authentication))
+        .then((response) => {
+          this.setState({
+            topics: _get(response, 'data.data.topics'),
+          })
+          if (_get(this.state, 'topics.length') > 0) {
```
NikkiMeganMoore, line 220: If you are checking for a state variable immediately after setting the state, it doesn't always behave like you think. It is better to set the initial topics state inside the constructor

src/components/DataMessenger/DataMessenger.js
```diff
@@ -207,6 +210,26 @@ export default class DataMessenger extends React.Component {
...
+            const topicsMessageContent = this.createTopicsMessage()
+            if (topicsMessageContent) {
+              this.state.messages.push(
+                this.createIntroMessage({ content: topicsMessageContent })
+              )
```
NikkiMeganMoore, line 225: We should never mutate the state directly (same goes for props. React is designed to be in full control of state and props). Whenever we want to change the state we should always use `setState` 

so something like
`const topicsMessage = this.createIntroMessage({ content: topicsMessageContent })
this.setState({
   messages: [
      ...this.state.messages,
      topicsMessage
   ]
})
`

NikkiMeganMoore (changes requested):

NikkiMeganMoore (approved):

PR #355, opened 2022-04-19
Ajax CSV Export

NikkiMeganMoore:


src/components/OptionsToolbar/OptionsToolbar.js
```diff
@@ -161,6 +168,33 @@ export default class Input extends React.Component {
       })
   }
 
+  onCSVMenuButtonClick = () => {
+    this.setState({ activeMenu: undefined })
+    const displayType = _get(this.props.responseRef, 'props.displayType')
+    const isPivotTable = displayType === 'pivot_table'
+
+    if (this.props.onCSVExportClick) {
```
AlexLiu-Chata, line 176: I am confused here...
`if (this.props.onCSVExportClick)` this one will always return "TRUE"

https://stackoverflow.com/questions/19839952/all-falsey-values-in-javascript
Some more truthy values
These are just a few values that some people might expect to be falsey, but are actually truthy.

-1 and all non-zero negative numbers

' ', " ", "false", 'null'... all non-empty strings, including strings that are just whitespace

Anything from typeof, which always returns a non-empty string, for example:

typeof null (returns a string 'object' [due to a longstanding bug/quirk](https://stackoverflow.com/questions/18808226/why-is-typeof-null-object))
typeof undefined (returns a string 'undefined')
Any object (except that "wilful violation" document.all in browsers). Remember that null isn't really an object, despite typeof suggesting otherwise. Examples:

{}
[]
function(){} or [() => {}](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions) (any function, including empty functions)
Error and any instance of Error
Any regular expression
Anything created with new (including new Number(0) and new Boolean(false))

src/js/queryService.js
```diff
@@ -247,6 +253,14 @@ export const exportCSV = ({ queryId, domain, apiKey, token } = {}) => {
       Authorization: `Bearer ${token}`,
     },
     responseType: 'blob',
+    onDownloadProgress: (progressEvent) => {
+      if (csvProgressCallback) {
```
AlexLiu-Chata, line 257: function(){} or [() => {}] (any function, including empty functions) will return true

src/components/OptionsToolbar/OptionsToolbar.js
NikkiMeganMoore: If the prop isn't provided then the value is undefined since there is no default prop here. Usually we do an empty function as a default but the default here is undefined

AlexLiu-Chata (approved):

PR #469, opened 2022-08-31
Ac 2223

alexzhou-chata:


src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -192,9 +189,32 @@ export default class ChataChart extends Component {
...
+    } catch (error) {
+      console.error(error)
+      return props.data
+    }
+  }
+
   aggregateRowData = (props) => {
     const { stringColumnIndex, numberColumnIndices } = props
-    const sortedData = _sortBy(props.data, (row) => row?.[stringColumnIndex])
+    const sortedData = this.sortChartDataByDate(props)
```
NikkiMeganMoore, line 214: This probably won't work the way you are expecting it to. The reason we have to sort this by `stringColumnIndex` is because when the aggregation loops through the data it compares the value to the previous row and if its the same it adds them together. The stringColumnIndex is the column that is on the chart axis, which is the column we want to aggregate by.

In other words, the chart will not always have the `DATE` column on the axis. Especially because we allow the user to choose a different column in the axis selector dropdown. Once the user selects a different ordinal axis, the data will have to be aggregated again by that new column.



Instead of changing the column to sort by, can you detect what type of column the `stringColumnIndex` is, and sort it according to that?

src/js/Constants.js
```diff
@@ -38,11 +38,11 @@ export const MONTH_NAMES = [
...
-  'Thursday',
-  'Friday',
   'Saturday',
+  'Friday',
+  'Thursday',
+  'Wednesday',
+  'Tuesday',
+  'Monday',
+  'Sunday',
 ]
```
NikkiMeganMoore, line 48: This is interesting I never thought to just reverse the order in here. Doesn't the table show up in the wrong order in this case?

src/js/Constants.js
```diff
@@ -38,11 +38,11 @@ export const MONTH_NAMES = [
...
-  'Wednesday',
-  'Thursday',
-  'Friday',
   'Saturday',
+  'Friday',
+  'Thursday',
+  'Wednesday',
+  'Tuesday',
+  'Monday',
+  'Sunday',
```
NikkiMeganMoore, line 47: This will need to be `Monday - Sunday` instead of `Sunday - Saturday`

NikkiMeganMoore (changes requested):

src/js/Constants.js
alexzhou-chata: Monday on the top.
I think it show up right

src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -192,9 +191,35 @@ export default class ChataChart extends Component {
     return { chartHeight, chartWidth }
   }
 
+  sortChartDataByDate = (props) => {
+    try {
+      if (!props.data || typeof props.data !== 'object') {
+        return undefined
+      }
+      const dateColumnIndex = getDateColumnIndex(props.columns)
```
NikkiMeganMoore, line 199: There may be more than 1 date column so "getDateColumnIndex" might not always find the right one (I think it just picks the first one)

Instead you could just grab props.stringColumnIndex since you already checked that it was a date type column below

NikkiMeganMoore (changes requested):

src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -23,17 +23,19 @@ import { ChataStackedLineChart } from '../ChataStackedLineChart'
...
   dataStructureChanged,
   getLegendLabelsForMultiSeries,
   getLegendLocation,
 } from '../helpers.js'
-
 import './ChataChart.scss'
-import { getColumnTypeAmounts } from '../../QueryOutput/columnHelpers'
+import {
+  getColumnTypeAmounts,
+  getDateColumnIndex,
```
NikkiMeganMoore, line 36: remove unused imports

src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -192,9 +191,35 @@ export default class ChataChart extends Component {
     return { chartHeight, chartWidth }
   }
 
+  sortChartDataByDate = (props) => {
+    try {
+      if (!props.data || typeof props.data !== 'object') {
+        return undefined
```
NikkiMeganMoore, line 197: Instead of returning undefined we should probably return the original `props.data` just to keep it the same going in as going out

src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -192,9 +191,35 @@ export default class ChataChart extends Component {
     return { chartHeight, chartWidth }
   }
 
+  sortChartDataByDate = (props) => {
+    try {
+      if (!props.data || typeof props.data !== 'object') {
```
NikkiMeganMoore, line 196: Is there a specific reason for these 2 checks and not just something like `props.data?.length`?

src/js/Constants.js
```diff
@@ -39,10 +39,10 @@ export const MONTH_NAMES = [
...
-  'Tuesday',
-  'Wednesday',
-  'Thursday',
-  'Friday',
   'Saturday',
+  'Friday',
+  'Thursday',
+  'Wednesday',
+  'Tuesday',
+  'Monday',
```
NikkiMeganMoore, line 47: LGTM 😂

NikkiMeganMoore (changes requested):

src/components/Charts/ChataChart/ChataChart.js
alexzhou-chata: removed

NikkiMeganMoore (approved):

PR #470, opened 2022-09-06
Query output project

NikkiMeganMoore:


src/components/QueryOutput/QueryOutput.js
```diff
@@ -0,0 +1,1988 @@
...
+      return (headerValue, rowValue, rowData, filterParams) => {
+        try {
+          const trimmedValue = headerValue.trim()
+          if (trimmedValue.length >= 2) {
+            const number = Number(
+              trimmedValue.substr(1).replace(/[^0-9.]/g, '')
+            )
+            if (trimmedValue[0] === '>' && trimmedValue[1] === '=') {
+              return rowValue >= number
+            } else if (trimmedValue[0] === '>') {
```
alexzhou-chata, line 1089: Line 1089 to 1098, maybe change to a Switch
[JavaScript: Switch vs. If Else](https://medium.com/@michellekwong2/switch-vs-if-else-1d458e7b0711#:~:text=The%20if%2Delse%20statement%20checks,character%20or%20an%20integer%20datatype.)

> Some key advantages of switch over if-else ladder:
> - A switch statement works much faster than an equivalent if-else ladder. It’s because the compiler generates a jump table for a switch during compilation. As a result, during execution, instead of checking which case is satisfied, it only decides which case has to be executed.
> - It’s more readable compared to if-else statements.

NikkiMeganMoore: Really good thought about improving efficiency. I think this function is one of the more important ones to improve efficiency on because it is iterated on all rows every time a table is filtered. 

I just looked into this and came across this stack overflow post which seems to contradict that article though which is interesting.. https://stackoverflow.com/questions/6665997/switch-statement-for-greater-than-less-than

Take a look and let me know what you think

src/components/QueryOutput/QueryOutput.js
```diff
@@ -0,0 +1,1988 @@
...
+
+    // Supported display types may have changed after initial data generation
+    this.initialSupportedDisplayTypes = this.getCurrentSupportedDisplayTypes()
+
+    this.state = {
+      displayType: this.getDisplayTypeFromInitial(props),
+      supportedDisplayTypes: this.initialSupportedDisplayTypes,
+      columns,
+      tableFilters: [],
+      selectedSuggestion: props.selectedSuggestion,
```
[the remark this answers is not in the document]
NikkiMeganMoore: I think the idea behind this was so you could provide an initial selection, but didn't have to be responsible for keeping the state in the parent component. Mainly because QueryOutput can be used in isolation and I wanted there to be as little work as possible for the person using the component. I am definitely open to changing this though if it makes more sense to control it by props only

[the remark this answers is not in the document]
NikkiMeganMoore: Actually, I don't think we need this to be a prop at all. I just looked and there is no other component that passes in the selectedSuggestion prop. 

It might have been there before because at some point maybe the QueryOutput component was unmounted (maybe someone changed the page) and the current selection got reset, so I wanted a way to set it back to the last selection. We could keep it as an optional prop in this case if someone wanted to set it but instead of `selectedSuggestion` we could call it `defaultSelectedSuggestion`? 

What do you think?

src/components/QueryOutput/QueryOutput.js
```diff
@@ -0,0 +1,1988 @@
...
+  }
+
+  updateColumns = (columns) => {
+    if (columns) {
+      const formattedColumns = this.formatColumnsForTable(columns)
+      this.setState({
+        columns: formattedColumns,
+        columnChangeCount: this.state.columnChangeCount + 1,
+      })
+      /* ----------- from merge conflict: uncomment this if there is a bug
```
Ben-CHATA, line 450: How old is this comment?

If it's been tested we're all clear, we should remove it

src/components/QueryOutput/QueryOutput.js
```diff
@@ -0,0 +1,1988 @@
...
+      // Regardless of the BE response, we want to default to percent
+      // if (
+      //   (col.type === 'RATIO' || col.type === 'NUMBER') &&
+      //   _get(
+      //     getDataFormatting(this.props.dataFormatting),
+      //     'comparisonDisplay'
+      //   ) === 'PERCENT'
+      // ) {
+      //   col.type = 'PERCENT'
+      // }
```
Ben-CHATA, line 1151: I think I left this. We can probably remove it unless you feel there might be a reason to keep?

src/components/QueryOutput/QueryOutput.js
```diff
@@ -0,0 +1,1988 @@
...
+        displayType: 'table',
+        supportedDisplayTypes: supportedDisplayTypes.filter(
+          (displayType) => displayType !== 'pivot_table'
+        ),
+      })
+    }
+  }
+
+  generatePivotTableData = ({ isFirstGeneration } = {}) => {
+    try {
```
Ben-CHATA, line 1331: CAn you explain why you used 'let' instead of 'const' for tableData, uniqueValues0, uniqueValues1 etc?

src/components/QueryOutput/QueryOutput.js
```diff
@@ -0,0 +1,1988 @@
...
+      return this.renderMessage(queryResponse.data)
+    }
+
+    // There is no error message in the response, display default error message
+    return this.renderMessage()
+  }
+
+  isDataLimited = () => {
+    const numRows = this.queryResponse?.data?.data?.rows?.length
+    const maxRowLimit = this.queryResponse?.data?.data?.row_limit
```
Ben-CHATA, line 1721: general question: does this method remove the need to use `_get`?

src/components/QueryOutput/QueryOutput.scss
```diff
@@ -0,0 +1,334 @@
...
+.react-autoql-query-validation-container {
+  width: 100%;
+}
+.react-autoql-query-validation-description {
+  margin-bottom: 14px;
+}
+
+.context-menu {
+  background: var(--react-autoql-background-color-secondary);
+  /* height: 100px; */
```
Ben-CHATA, line 260: remove

src/components/QueryOutput/QueryOutput.test.js
```diff
@@ -0,0 +1,108 @@
...
+  const setupProps = { ...defaultProps, ...props }
+  const wrapper = shallow(<QueryOutput {...setupProps} />).dive()
+  if (state) {
+    wrapper.setState(state)
+  }
+  return wrapper
+}
+
+describe('test each response case', () => {
+  for (let i = 0; i < testCases.length; i++) {
```
Ben-CHATA, line 21: Cool use of a for loop inside of a testing suite!

src/components/QueryOutput/QueryOutput.test.js
```diff
@@ -0,0 +1,108 @@
...
+
+    // describe('props', () => {
+    //   test('does not throw warning with expected props', () => {
+    //     checkProps(QueryOutput, defaultProps)
+    //   })
+    //   describe('showMask', () => {
+    //     test('showMask false does not show mask on drawer open', () => {})
+    //     test('showMask true shows mask on drawer open', () => {})
+    //   })
+    // })
```
Ben-CHATA, line 49: I was reading up on some guiding principles of FE unit testing and apparently tests like these aren't actually required or really all that useful.

according to [this doc](https://testing-library.com/docs/guiding-principles/), and others like it, we should be testing how the user interacts with the UI, rather than it's underlying functionality.

src/components/QueryOutput/QueryOutput.js
```diff
@@ -0,0 +1,1988 @@
...
+      this.tableRef.copyToClipboard()
+    } else if (
+      this.state.displayType === 'pivot_table' &&
+      this.pivotTableRef?._isMounted
+    ) {
+      this.pivotTableRef.copyToClipboard()
+    }
+  }
+
+  getBase64Data = () => {
```
Ben-CHATA, line 643: I'm wondering if this could actually be a helper function?

Shiv-Chata:
This is a nicely done component, well done. I have a couple of question and thoughts i'd like to share tho. 

Instead of using and importing _get for lodash, the optional chaining operator should be able to replace it. As well this operator can be used for indexing as well like blah?.[index], as an extra safety precaution. However, there is a risk of overusing this operator.  

I noticed many console.log()'s through the component. Is keeping console logs around, even for just errors, something we want to do outside the environment. Is there a risk of providing too much info from the logs? 

Would it be a good idea to separate this component up a it more to be specific to the parents use for it? I noticed the props being passed into the component vary a decent amount based on the parent...

Have you considered making the transition from class to functional components?

src/components/QueryOutput/QueryOutput.js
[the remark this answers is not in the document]
AlexLiu-Chata: > I think the idea behind this was so you could provide an initial selection, but didn't have to be responsible for keeping the state in the parent component. Mainly because QueryOutput can be used in isolation and I wanted there to be as little work as possible for the person using the component. I am definitely open to changing this though if it makes more sense to control it by props only

Then, It makes sense. maybe just add some comments here and let others know why we are using props as initial state.

PR #499, opened 2022-10-13
feat: 🎸 Cancel query when user delete message if it's not done

alexzhou-chata:
Cancel query if user deletes message before request is finished

✅ Closes: AC-2398


src/components/ChatContent/ChatContent.js
```diff
@@ -270,14 +285,19 @@ export default class ChatContent extends React.Component {
...
+                target="_blank"
+                rel="noreferrer"
+              >
                 Click here to authorize access then try querying again.
               </a>
             </span>
           ),
         })
-      } else {
+      } else if (_get(response?.data, 'message') !== 'Request cancelled') {
```
NikkiMeganMoore, line 300: Can you use optional chaining instead? My goal is to eventually get rid of all _get functions eventually

src/js/queryService.js
```diff
@@ -262,7 +262,9 @@ export const runQuery = (params) => {
         return runQueryOnly(params)
       })
       .catch((error) => {
-        console.error(error)
+        if (_get(error?.data, 'message') !== 'Request cancelled') {
```
NikkiMeganMoore, line 265: optional chaining here as well please

NikkiMeganMoore, line 265: Please use the `CANCELLED` enum here

src/components/ChatContent/ChatContent.js
```diff
@@ -420,6 +448,9 @@ export default class ChatContent extends React.Component {
             dataPageSize={this.props.dataPageSize}
             isResizing={this.props.isResizing}
             shouldRender={this.props.shouldRender}
+            setAxiosSource={(axiosSource) => {
```
NikkiMeganMoore, line 451: Instead of using a callback to set another axios source, you can just call `queryInputRef.cancelQuery()` and it should handle everything for you

src/components/ChatContent/ChatContent.js
NikkiMeganMoore, line 300: Use `CANCELLED` enum here as well please

src/components/ChatContent/ChatContent.js
```diff
@@ -251,6 +263,9 @@ export default class ChatContent extends React.Component {
       this.addMessage(message)
     }
   }
+  cancelAllQueries = () => {
```
NikkiMeganMoore, line 266: Probably don't need this function if you just use QueryInputs function

NikkiMeganMoore (changes requested):

NikkiMeganMoore (approved):

PR #502, opened 2022-10-13
feat: 🎸 add row count dropdown

alexzhou-chata:
✅ Closes: AC-2131


src/components/Charts/Axes/Axes.js
```diff
@@ -1,28 +1,34 @@
...
-
+    this.COMPONENT_ID = uuid()
     this.xAxisKey = uuid()
     this.yAxisKey = uuid()
     this.axisLabelPaddingTop = 5
     this.axisLabelPaddingLeft = 10
-
+    this.maxRows = 5000
+    this.initialRowNumber = 50
+    this.visualizingSpan = document.getElementById('visualizing-span')
```
NikkiMeganMoore, line 20: Just to confirm, this constructor is only called once before it mounts. If visualizing-span is rendered in this component, then it won't exist at this time.

Also we probably want to add a unique identifier to the ID `visualizing-span`. If you are in a dashboard it will take the first element it finds, which could be in a different QueryOutput

src/components/Charts/Axes/Axes.js
```diff
@@ -36,7 +42,23 @@ export default class Axes extends React.Component {
...
+  componentDidUpdate = (prevProps, prevState) => {
+    if (prevState.currentRowNumber !== this.state.currentRowNumber) {
+      this.forceUpdate()
+    }
+  }
 
+  renderLoadMoreDropdown = (currentRowNumber, totalRowNumber) => {
+    return (
+      <tspan id={`load-more-drop-down-span-${this.COMPONENT_ID}`}>
+        <tspan id={`visualizing-span-${this.COMPONENT_ID}`}>{`Visualizing `}</tspan>
```
NikkiMeganMoore, line 54: I see the unique identifier here.. should you remove that variable you set in the constructor?

NikkiMeganMoore: ideally we should use refs as much as we can instead of document.querySelector. We can make a jira for this to do later though as this feature is high priority for nike

NikkiMeganMoore (changes requested):

NikkiMeganMoore (approved):

PR #530, opened 2022-11-08
Scale on chart

alexzhou-chata:


src/components/Charts/Axis/AxisScaler.js
```diff
@@ -0,0 +1,29 @@
...
+    return (
+      <rect
+        {...this.props.childProps}
+        className='axis-scaler-border'
+        data-test='axis-scaler-border'
+        fill='transparent'
+        stroke='transparent'
+        strokeWidth='1px'
+        rx='4'
+        onClick={() => {
```
NikkiMeganMoore, line 23: This can just be `onClick={this.props.setIsChartScaled}`

NikkiMeganMoore (changes requested):

PR #535, opened 2022-11-15
Ac 2480

alexzhou-chata:


AlexLiu-Chata (dismissed):

src/components/QueryOutput/QueryOutput.js
```diff
@@ -1976,9 +1978,12 @@ export class QueryOutput extends React.Component {
   renderFooter = () => {
     const shouldRenderRT = this.shouldRenderReverseTranslation()
     const shouldRenderDLW = this.shouldRenderDataLimitWarning()
-
+    const footerClassName =
```
NikkiMeganMoore, line 1981: This seems a bit confusing to me. What if we did something like this

```
const footerClassName =
    `query-output-footer-top
    ${!shouldRenderRT ? 'no-margin' : ''}
    ${this.props.reverseTranslationPlacement}`
```

Then the css selector could just be

`.query-output-footer.top` or `.query-output-footer.bottom`

NikkiMeganMoore (changes requested):

NikkiMeganMoore (approved):

PR #585, opened 2023-02-24
feat: 🎸 delete create alert button

alexzhou-chata:
✅ Closes: AC-2611

### DIFF
--- a/src/components/Notifications/DataAlerts/DataAlerts.js
+++ b/src/components/Notifications/DataAlerts/DataAlerts.js
@@ -5,7 +5,7 @@ import _get from 'lodash.get'
 import _isEqual from 'lodash.isequal'
 import _cloneDeep from 'lodash.clonedeep'
 import { v4 as uuid } from 'uuid'
-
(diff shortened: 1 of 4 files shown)


src/components/Notifications/DataAlerts/DataAlerts.js
```diff
@@ -215,8 +197,9 @@ class DataAlerts extends React.Component {
 
     return (
       <div className='data-alerts-list-container'>
-        {type === 'custom' &&
-          this.renderNotificationGroupTitle('Custom Data Alerts', 'View and manage your custom Data Alerts', true)}
+        {type === 'custom' && _get(list, 'length')
```
NikkiMeganMoore, line 200: We could probably just use list?.length instead of lodash here

NikkiMeganMoore (dismissed):

src/components/Notifications/DataAlerts/DataAlerts.js
alexzhou-chata: yep, for sure!

NikkiMeganMoore (approved):

PR #589, opened 2023-03-02
Project march2

NikkiMeganMoore:


src/components/Charts/Bars/Bars.js
```diff
@@ -0,0 +1,102 @@
...
+  static propTypes = chartElementPropTypes
+  static defaultProps = chartElementDefaultProps
+
+  state = {
+    activeKey: this.props.activeChartElementKey,
+  }
+
+  shouldComponentUpdate = () => {
+    return true
+  }
```
Ben-CHATA, line 15: how does this work? lol

Ben-CHATA (approved):

src/components/Charts/Bars/Bars.js
NikkiMeganMoore: This actually used to have more in it then got removed so I can take this out.

src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -0,0 +1,531 @@
...
+  getInnerDimensions = () => {
+    const { chartWidth, chartHeight } = this.getRenderedChartDimensions()
+
+    const propsWidth = typeof this.props.width === 'number' ? this.props.width : undefined
+    const propsHeight = typeof this.props.height === 'number' ? this.props.height : undefined
+
+    const containerWidth = propsWidth ?? this.chartContainerRef?.clientWidth ?? 0
+    const containerHeight = propsHeight ?? this.chartContainerRef?.clientHeight ?? 0
+
+    let innerWidth = containerWidth - 2 * this.PADDING
```
Shiv-Chata, line 273: I noticed the value 2 is used a couple times around here, would it be worth it to also make it a variable like this.PADDING?

src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -0,0 +1,531 @@
...
+    if (this.innerChartRef?.xScale && chartWidth) {
+      const rangeInPx = this.innerChartRef.xScale.range()[1] - this.innerChartRef.xScale.range()[0]
```
Shiv-Chata, line 275: Does range return the upper and lower bound for the xScale? Didn't know js had a range function.

src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    let innerHeight = containerHeight - 2 * this.PADDING
+    if (this.innerChartRef?.yScale && chartHeight) {
+      const rangeInPx = this.innerChartRef.yScale.range()[0] - this.innerChartRef.yScale.range()[1]
+      const totalVerticalMargins = chartHeight - rangeInPx
+      innerHeight = containerHeight - totalVerticalMargins - 2 * this.PADDING
+    }
+
+    if (innerWidth < 1) {
```
Shiv-Chata, line 287: Could you Ceiling this value when it is calculated above? Or should some values have decimals?

src/components/Charts/ChataChart/ChataChart.js
```diff
@@ -0,0 +1,531 @@
...
+
+  saveAsPNG = () => {
+    const svgElement = this.chartRef
+    if (!svgElement) {
+      return
+    }
+
+    svgToPng(svgElement, 20)
+      .then((data) => {
+        let dt = data // << this fails in IE/Edge...
```
Shiv-Chata, line 353: Why does this fail in IE/Edge?

src/components/Charts/ChataPieChart/ChataPieChart.js
```diff
@@ -0,0 +1,360 @@
...
+
+export default class ChataPieChart extends Component {
+  constructor(props) {
+    super(props)
+
+    this.CHART_ID = uuid()
+    this.LEGEND_ID = `react-autoql-pie-legend-${uuid()}`
+
+    this.sortedData = props.data
+      .concat() // this copies the array so the original isn't mutated
```
Shiv-Chata, line 26: Nice trick! I think recently the structuredClone() is now supported by all the main browsers, so this might be a good alternative to use.

PR #665, opened 2023-06-28
try to fix AC-2942

alexzhou-chata:


src/components/Notifications/RuleSimple/RuleSimple.js
```diff
@@ -201,7 +201,21 @@ export default class RuleSimple extends React.Component {
 
     return expression
   }
+  isIsoDate = (str) => {
```
NikkiMeganMoore, line 204: doesn't this function already exist somewhere else?

NikkiMeganMoore (changes requested):

NikkiMeganMoore (approved):

PR #685, opened 2023-07-19
fixed the bug

alexzhou-chata:


src/components/Charts/ChataChart/ChataChart.scss
```diff
@@ -8,6 +8,9 @@
 
   .react-autoql-slider-wrapper {
     font-size: var(--react-autoql-chart-font-size, 12px);
+	&.mobile{
+		font-size: 9px;
```
Ben-CHATA, line 12: px is slightly different on all native devices.

I think collectively we need to start considering using REM more instead of PX for things like font size, padding, margins etc. But this is particularly important for mobile.

https://www.linkedin.com/pulse/how-using-rem-can-help-you-build-excellent-andr%C3%A9-fillype/

Ben-CHATA (changes requested):

src/components/Charts/ChataChart/ChataChart.scss
alexzhou-chata: > px is slightly different on all native devices.
> 
> I think collectively we need to start considering using REM more instead of PX for things like font size, padding, margins etc. But this is particularly important for mobile.
> 
> https://www.linkedin.com/pulse/how-using-rem-can-help-you-build-excellent-andr%C3%A9-fillype/

Agree with you

Ben-CHATA (approved):

PR #695, opened 2023-09-19
capture data object when user change axes

alexzhou-chata:


src/components/Charts/Axis/Axis.js
```diff
@@ -91,6 +91,32 @@ export default class Axis extends Component {
   }
 
   componentDidUpdate = (prevProps, prevState) => {
+    if (
+      (!_isEqual(prevProps.visibleSeriesIndices, this.props.visibleSeriesIndices) ||
+        prevProps.stringColumnIndex !== this.props.stringColumnIndex) &&
+      this.state.isAxisSelectorOpen !== prevState.isAxisSelectorOpen
```
NikkiMeganMoore, line 97: Would it be better/easier to put this stuff in `onChangeNumberColumnIndices` and `on ChangeStringColumnIndex`? Instead of checking every state update? If you have thought about this and did this intentionally then that's fine. Just making sure

alexzhou-chata: Sorry I wasn't aware of those two callback function, I will transfer it to there. Thx for your advice!

NikkiMeganMoore (approved):

PR #705, opened 2023-09-29
Remove references to  in Data Alerts and Notifications

alexzhou-chata:


src/components/Notifications/DataAlertModal/DataAlertModal.js
```diff
@@ -516,7 +510,6 @@ class DataAlertModal extends React.Component {
           messageInput={this.state.messageInput}
           onTitleInputChange={(e) => this.setState({ titleInput: e.target.value })}
           onMessageInputChange={(e) => this.setState({ messageInput: e.target.value })}
-          queryText={this.getQueryText()}
```
NikkiMeganMoore, line 519: We still need to keep this I think. We can display the query if `this.props.queryResponse?.data?.data?.text` exists

src/components/Notifications/DataAlertModal/DataAlertModal.js
```diff
@@ -654,13 +647,7 @@ class DataAlertModal extends React.Component {
     return <span key={`title-icon-${this.COMPONENT_KEY}`} />
   }
 
-  getQueryText = () => {
-    return this.props.queryResponse?.data?.data?.text ?? this.props.currentDataAlert?.data_return_query
```
NikkiMeganMoore, line 658: here we can just return `this.props.queryResponse?.data?.data?.text`

NikkiMeganMoore (changes requested):

NikkiMeganMoore (approved):

PR #708, opened 2023-10-04
Handle case where data return is null in a notification

alexzhou-chata:


src/components/Notifications/NotificationQueryResponse/NotificationQueryResponse.js
Shiv-Chata, line 1: Should a message be added to let the user know there wasnt a response? Instead of just having loading dots for this case

Shiv-Chata (approved):

PR #727, opened 2023-10-20
add agg query vs single value UI

alexzhou-chata:


src/components/Notifications/RuleSimple/RuleSimple.js
```diff
@@ -196,6 +206,12 @@ export default class RuleSimple extends React.Component {
         session_filter_locks: lockedFilters,
       },
     ]
+    const selectedNumberColumnName = this.state.selectedColumns.map(
+      (index) => this.props.queryResponse?.data?.data?.columns[index]?.name,
+    )[0]
+    if (this.NUMBER_COLUMNS_AMOUNT >= 2) {
```
NikkiMeganMoore, line 212: Looking at this I am thinking we should have a condition to exclude the groupables from this number columns amount variable. There is a case where there could be 2 number columns, but one of them is a groupable, so really only 1 number column would be selectable

src/components/Notifications/ConditionBuilder/ConditionBuilder.js
```diff
@@ -28,6 +28,7 @@ export default class ConditionBuilder extends React.Component {
     conditionStatementOnly: PropTypes.bool, // Set this to true if you want a summary of the expression without needing to interact with it
     onChange: PropTypes.func, // this returns 2 params (isSectionComplete, expressionJSON)
     onLastInputEnterPress: PropTypes.func,
+    dataFormatting: dataFormattingType.isRequired,
```
NikkiMeganMoore, line 31: Shouldn't be required. We can use the default values if its not provided

src/components/Notifications/DataAlertModal/DataAlertModal.js
```diff
@@ -66,6 +67,7 @@ class DataAlertModal extends React.Component {
     onClosed: PropTypes.func,
     onOpened: PropTypes.func,
     editView: PropTypes.bool,
+    dataFormatting: dataFormattingType.isRequired,
```
NikkiMeganMoore, line 70: same as above

src/components/Notifications/RuleSimple/RuleSimple.js
```diff
@@ -95,6 +105,7 @@ export default class RuleSimple extends React.Component {
     initialData: PropTypes.arrayOf(PropTypes.shape({})),
     queryResponse: PropTypes.shape({}),
     onLastInputEnterPress: PropTypes.func,
+    dataFormatting: dataFormattingType.isRequired,
```
NikkiMeganMoore, line 108: same

src/components/OptionsToolbar/OptionsToolbar.js
```diff
@@ -51,7 +52,7 @@ export class OptionsToolbar extends React.Component {
   static propTypes = {
     authentication: authenticationType,
     autoQLConfig: autoQLConfigType,
-
+    dataFormatting: dataFormattingType.isRequired,
```
NikkiMeganMoore, line 55: same

NikkiMeganMoore (changes requested):

NikkiMeganMoore (approved):

PR #765, opened 2023-11-23
add upload training data when user switch axes

alexzhou-chata:


src/components/QueryOutput/QueryOutput.js
```diff
@@ -2313,6 +2419,9 @@ export class QueryOutput extends React.Component {
           queryFn={this.queryFn}
           onBucketSizeChange={this.props.onBucketSizeChange}
           bucketSize={this.props.bucketSize}
+          queryResponse={this.queryResponse}
```
NikkiMeganMoore, line 2422: Do we need to pass this in anymore? It is a ton of data and might slow down performance if its not needed

src/components/QueryOutput/QueryOutput.js
```diff
@@ -851,7 +877,87 @@ export class QueryOutput extends React.Component {
 
     return undefined
   }
-
+  getTrainingData = () => {
+    const isYScaleStringColumn = !this.chartRef?.innerChartRef?.yScale.fields[0].hasOwnProperty('aggType')
```
NikkiMeganMoore, line 881: We actually have a state variable in here (QueryOutput) called `aggConfig` where all these agg types are stored when they change. It would be simpler to just grab it from there I think

src/components/QueryOutput/QueryOutput.js
```diff
@@ -851,7 +877,87 @@ export class QueryOutput extends React.Component {
...
+    const isXScaleStringColumn = !this.chartRef?.innerChartRef?.xScale.fields[0].hasOwnProperty('aggType')
+    const isTableFilterInXScale = this.queryResponse?.data?.data?.fe_req.filters.some(
+      (obj) => obj.name === this.chartRef?.innerChartRef?.xScale.column.name,
+    )
+    const isTableFilterInYScale = this.queryResponse?.data?.data?.fe_req.filters.some(
+      (obj) => obj.name === this.chartRef?.innerChartRef?.yScale.column.name,
+    )
```
NikkiMeganMoore, line 888: What if there is a second y axis? ie. a column_line combo chart. Or a pie chart with no axes?

src/components/QueryOutput/QueryOutput.js
```diff
@@ -851,7 +877,87 @@ export class QueryOutput extends React.Component {
 
     return undefined
   }
-
+  getTrainingData = () => {
```
NikkiMeganMoore, line 880: I think ideally we will want to move this function outside of the component, and pass in any required data through function params. We will want to duplicate this functionality in vanilla so the more we can separate the better

NikkiMeganMoore (changes requested):

src/components/QueryOutput/QueryOutput.js
alexzhou-chata: nope, I forgot to delete it, thank you for reminding me!

PR #768, opened 2023-12-06
Ac 3164

alexzhou-chata:
fix data alert notification bug

### DIFF
--- a/src/components/ChataTable/ChataTable.js
+++ b/src/components/ChataTable/ChataTable.js
@@ -44,15 +44,14 @@ export default class ChataTable extends React.Component {
     if (isNaN(this.totalPages) || !this.totalPages) {
       this.totalPages = 1
     }
+    if (props.initialParams) {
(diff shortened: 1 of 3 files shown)


Shiv-Chata (approved):

src/components/ChataTable/ChataTable.js
```diff
@@ -44,17 +44,16 @@ export default class ChataTable extends React.Component {
     if (isNaN(this.totalPages) || !this.totalPages) {
       this.totalPages = 1
     }
+    if (props.initialParams) {
+      this.tableParams = _cloneDeep(props.initialParams)
+    }
 
     this.tableParams = {
```
NikkiMeganMoore, line 51: What is the point of the above line if it just gets overwritten here?

NikkiMeganMoore (changes requested):

NikkiMeganMoore (approved):

PR #810, opened 2024-03-08
Custom columns

NikkiMeganMoore:


alexzhou-chata (approved):

src/components/AddColumnBtn/CustomColumnModal.js
```diff
@@ -0,0 +1,958 @@
...
+      <span>
+        <span>{WINDOW_FUNCTIONS[chunk.fn].label}( </span>
+        <Select
+          key={`custom-column-select-${i}`}
+          placeholder='Select a Column'
+          value={chunk.column?.field}
+          outlined={false}
+          showArrow={false}
+          className='react-autoql-available-column-selector'
+          // onChange={(value) => this.changeChunkValue(value, chunk.type, i)}
```
Shiv-Chata, line 553: Doesn't the Select need the onChange?

src/components/AddColumnBtn/CustomColumnModal.js
```diff
@@ -0,0 +1,958 @@
...
+          <>
+            <span>, Grouped by </span>
+            <Select
+              key={`custom-column-select-${i}`}
+              placeholder='Select a Column'
+              value={chunk.groupby}
+              outlined={false}
+              showArrow={false}
+              className='react-autoql-available-column-selector'
+              // onChange={(value) => this.changeChunkValue(value, chunk.type, i)}
```
Shiv-Chata, line 572: Same as above

src/components/AddColumnBtn/CustomColumnModal.js
```diff
@@ -0,0 +1,958 @@
...
+          <>
+            <span>, Ordered by </span>
+            <Select
+              key={`custom-column-select-${i}`}
+              placeholder='Select a Column'
+              value={chunk.orderby}
+              outlined={false}
+              showArrow={false}
+              className='react-autoql-available-column-selector'
+              // onChange={(value) => this.changeChunkValue(value, chunk.type, i)}
```
Shiv-Chata, line 593: Same as above

src/components/AddColumnBtn/customColumnHelpers.js
```diff
@@ -0,0 +1,187 @@
...
+export const OPERATORS = {
+  CONCAT: {
+    value: 'CONCAT',
+    label: 'Concatenate(...)',
+    js: '',
+  },
+  ADDITION: {
+    value: 'ADD',
+    label: <Icon type='plus' />,
+    icon: 'plus',
```
Shiv-Chata, line 16: label and icon values seem mismatched here... same with the below, so i am assuming this is intentional?

src/components/ChataTable/ChataTable.js
```diff
@@ -672,6 +697,17 @@ export default class ChataTable extends React.Component {
   //   this.ref?.tabulator?.clearHeaderFilter()
```
Shiv-Chata, line 697: Remove?

Shiv-Chata (commented):
Looks good, just left a couple small comments

src/components/AddColumnBtn/CustomColumnModal.js
```diff
@@ -0,0 +1,958 @@
...
+
+const globalOperators = ['LEFT_BRACKET', 'RIGHT_BRACKET']
+const FUNCTION_OPERATOR = 'FUNCTION'
+
+const HIGHLIGHTED_CLASS = 'highlighted-column'
+const DISABLED_CLASS = 'disabled-column'
+const FORMULA_CLASS = 'formula-column'
+const DEFAULT_COLUMN_NAME = 'New Column'
+
+export const getSelectableColumns = (columns) => {
```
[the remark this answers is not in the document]
NikkiMeganMoore: Ohh yes I actually did this intentionally, because the getSelectableColumns function will eventually include more than just numerical columns as we add to the feature. So just future proofing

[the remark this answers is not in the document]
NikkiMeganMoore: Yes you are probably right. I will do that

src/components/AddColumnBtn/customColumnHelpers.js
NikkiMeganMoore: Yes it is intentionally but I'm actually changing this to either have icon OR label. If icon exists, then the react widgets will use it for the <Icon /> type, and if not, it will use the label as plain text.

I put this into autoql-fe-utils now without the <Icon/> part to make it compatible with both react and vanilla

src/components/ChataTable/ChataTable.js
NikkiMeganMoore: See comment above this line. Just want to keep this for future feature of "clear all filters" button. It needs a little more work and not a huge priority for now

Shiv-Chata (approved):

PR #860, opened 2024-07-22
Ant 2710

alexzhou-chata:


src/components/QueryOutput/QueryOutput.js
```diff
@@ -923,7 +923,11 @@ export class QueryOutput extends React.Component {
           response = this.getNewResponseWithCustomColumns(response, this.state.customColumns)
         }
       } catch (error) {
-        response = error
+        if (error?.data?.message === REQUEST_CANCELLED_ERROR) {
```
Shiv-Chata, line 926: It would be good to create a `handleError` function for this logic, as i see it is used below as well.

Shiv-Chata (commented):
Left a comment

Shiv-Chata (approved):

PR #861, opened 2024-07-29
Ant 2690 -Delete all notifications for user

alexzhou-chata:


src/components/Notifications/NotificationFeed/NotificationFeed.js
```diff
@@ -508,12 +525,27 @@ class NotificationFeed extends React.Component {
...
 
   renderDeleteAllButton = () => {
-    return null
-    // Enable this once the batch delete endpoint is ready
+    let title = 'Delete all notifications?'
+    if (this.props.enableFetchAllNotificationFeedAcrossProjects) {
+      title = `Delete all notifications for ${this.props.selectedProjectName}?`
+    } else {
+      return
+      //Enable this button for alpha alerts only temporarily
```
Shiv-Chata, line 533: I don't think this comment makes sense to have here anymore now that the button is being rendered. 

One change you could do here is instead of having an empty return in the else block, at the top of the function you can do 
`if (!this.props.enableFetchAllNotificationFeedAcrossProjects) {
return 
}

// render button logic after...
`

src/components/Notifications/NotificationFeed/NotificationFeed.js
```diff
@@ -350,7 +351,23 @@ class NotificationFeed extends React.Component {
...
+    if (pagination) {
+      pagination = {
+        ...pagination,
+        total_items: pagination.total_items > 0 ? pagination.total_items - 1 : 0,
+      }
+    }
+
+    this.setState({
+      notificationList: newList,
+      pagination,
```
Shiv-Chata, line 366: If pagination is 0 will any notification be deleted?

src/components/Notifications/NotificationFeed/NotificationFeed.js
```diff
@@ -289,13 +294,10 @@ class NotificationFeed extends React.Component {
...
   }
   filterNotificationsByProject = () => {
-    const limit =
-      (this.state.notificationList?.length || this.NOTIFICATION_FETCH_LIMIT) + (this.state.unFetchedNotifications ?? 0)
-
     fetchNotificationFeed({
       ...getAuthentication(this.props.authentication),
       offset: 0,
-      limit,
+      limit: 10,
```
Shiv-Chata, line 300: Why is the limit defaulting to 10 here?

src/components/Notifications/NotificationFeed/NotificationFeed.js
```diff
@@ -262,7 +265,9 @@ class NotificationFeed extends React.Component {
     fetchNotificationFeed({
       ...getAuthentication(this.props.authentication),
       offset: 0,
-      limit,
+      limit: this.props.enableFetchAllNotificationFeedAcrossProjects ? 10 : limit,
```
Shiv-Chata, line 268: Could add this condition to where the limit variable is being defined above, so that all the logic for this `limit` field is in one spot/line

Shiv-Chata (commented):
Left a couple comments

src/components/Notifications/NotificationFeed/NotificationFeed.js
alexzhou-chata: The value of pagination won't affect the deleteAllNotifications() endpoint, but the UI/UX.
Pagination can only be an empty object rather than 0.
If pagination is {}, notifications will get deleted as well.

src/components/Notifications/NotificationFeed/NotificationFeed.js
alexzhou-chata: I will change it to the this.NOTIFICATION_FETCH_LIMIT

src/components/Notifications/NotificationFeed/NotificationFeed.js
alexzhou-chata: sure, good caught, thx mate!

Shiv-Chata (approved):

PR #865, opened 2024-08-07
modify css and add a successful callback

alexzhou-chata:
1. Modified a css to accommodate button squeezes between each other in a really narrow screen scenario, such as Galaxy Z Fold 5. 
Issue
Issue Fixed
2 Added successful callback to two delete notifications function that allow user to know the notifications have been deleted successfully.

Screenshot 2024-08-07 at 1 20 22 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "react-autoql",
-  "version": "8.6.2",
+  "version": "8.6.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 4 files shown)


Shiv-Chata:
The code looks good but for 1 the buttons do look squished. Have you considered using a menu icon that when a user clicks pops up a menu which hold all three of those buttons? With a menu, you will save a lot more space at the top of the list and it will also be easier to add new buttons in the future

Shiv-Chata (approved):

PR #867, opened 2024-08-26
modify UI for webapp

alexzhou-chata:


Shiv-Chata (commented):
I am a little concerned about using all those tertiary statements to define the className for some of these component. Is there a way to determine if `isMobile` is true in the css files and then putting the logic for if it is or is not mobile in there?

Shiv-Chata (approved):

PR #869, opened 2024-08-28
Ant 2812 add project selector to data messenger component 

Shiv-Chata:
- update autoQlDefaultConfig prop by adding enableProjectSelect value
- if enableProjectSelect is true then three other props that have been added become useful: 
  1.  projectSelectList,
  2. selectedProjectId 
  3. onProjectSelectChange
- the above three props handle the project selector in the DM component by allowing a list of projects which contains only the project id and project display name per list item, the selected project id is the current project, and onProjectSelectChange is the callback used to handle changing the project


src/components/DataMessenger/DataMessenger.js
```diff
@@ -251,7 +267,7 @@ export class DataMessenger extends React.Component {
 
       clearTimeout(this.windowResizeTimer)
       clearTimeout(this.executeQueryTimeout)
-    } catch (error) {}
+    } catch (error) { }
```
alexzhou-chata, line 270: maybe add a console.error ?

Shiv-Chata: I think since this is a clean up function there isn't really a need to console the error as it won't give useful information

alexzhou-chata (approved):

PR #874, opened 2024-09-03
Ant 2816 persist custom columns

Shiv-Chata:
- update queryOutput, chataTable, customColumnModal, RuleSimple, and Data Alert Modal components to allow for custom columns to be created, generated and persisted as expected with BE support. 
- Persist the custom columns
- Ability to create alerts based on custom columns 
- Ability to delete
- Ability to create a custom column from other custom columns

### DIFF
--- a/src/components/QueryOutput/QueryOutput.js
+++ b/src/components/QueryOutput/QueryOutput.js
@@ -477,7 +477,9 @@ export class QueryOutput extends React.Component {
       if (referenceIdNumber >= 200 && referenceIdNumber < 300) {
         return false
       }
-    } catch (error) { }
+    } catch (error) {
+      console.error(error)
+    }
     return true
   }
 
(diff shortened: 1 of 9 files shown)


src/components/QueryOutput/QueryOutput.js
```diff
@@ -581,7 +477,7 @@ export class QueryOutput extends React.Component {
       if (referenceIdNumber >= 200 && referenceIdNumber < 300) {
         return false
       }
-    } catch (error) {}
+    } catch (error) { }
```
alexzhou-chata, line 480: may need add something here? or, this catch will actually do nothing.

Shiv-Chata: Updated

alexzhou-chata (approved):
LGTM

PR #879, opened 2024-09-06
ANT-2787

eleon-chata-ai:
- update onRemoveColumnClick so that the columns are available to be reslected after removed from table

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-autoql",
-  "version": "8.8.10",
+  "version": "8.8.11",
   "description": "React Widget Library",
   "main": "dist/autoql.cjs.js",
   "module": "dist/autoql.esm.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-autoql",
-  "version": "8.7.2",
+  "version": "8.7.3",
   "description": "React Widget Library",
   "main": "dist/autoql.cjs.js",
   "module": "dist/autoql.esm.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-autoql",
-  "version": "8.7.3",
+  "version": "8.7.8",
   "description": "React Widget Library",
   "main": "dist/autoql.cjs.js",
   "module": "dist/autoql.esm.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-autoql",
-  "version": "8.8.8",
+  "version": "8.8.9",
   "description": "React Widget Library",
   "main": "dist/autoql.cjs.js",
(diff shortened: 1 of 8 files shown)


alexzhou-chata (dismissed):
LGTM

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-autoql",
-  "version": "8.8.7",
+  "version": "8.8.8",
```
alexzhou-chata, line 3: We had version 8.8.8 just now.
Should update to 8.8.9

eleon-chata-ai: updated to 8.8.9

alexzhou-chata (dismissed):

src/components/ChataTable/ChataTable.js
```diff
@@ -1068,11 +1068,18 @@ export default class ChataTable extends React.Component {
...
-        return select?.columns[0]?.replace(/ /g, '') !== column?.name?.replace(/ /g, '')
-      },
-    )
+    const newAdditionalSelectColumns = currentAdditionalSelectColumns?.map((select) => {
+      if (
+        select.columns[0] !== column.name ||
+        select?.columns[0]?.replace(/ /g, '') !== column?.name?.replace(/ /g, '')
+      ) {
+        return select
+      } else if (select.columns[0] === column.name) {
```
Shiv-Chata, line 1077: I think the .filter() still needs to be here, as `newAdditionalSelectColumns` get's passed into an endpoint that updates the query response and the .map() func doesn't remove items.

What is the goal here? The columns were being removed correctly but then were not showing up in the add column list?

Shiv-Chata (commented):
Left a comment

src/components/ChataTable/ChataTable.js
eleon-chata-ai: I wasn't able to use filter here. The filter was causing the column name to be removed from the list after it was being removed. Yes column was being removed, but was not in the add column list

Shiv-Chata: We need to filter here because we need to update the queryOutput at the sql level. By removing it from the additional_selects we update the query at the SQL level. 

If we check the query endpoint after removing the column we should see the removed column not included in the `additional_selects` lists of the payload, but then it should be in the response in the `available_selects` list. If it is in the available_selects list then something isn't updating properly, probably something in the AddColumnBtn component, if it isn't in the list then i think there might a bug in the BE.

eleon-chata-ai: i think the filter / map are just creating new array of values, which is still being saved to the variable. I changed because the filter is removing the option from the newAdditionalSelectColumns, vs hiding is using the visible flags. I think we should be using the following to hide the columns instead
column.visible = false
column.is_visible = false

Shiv-Chata (approved):
Lgtm!

PR #885, opened 2024-09-10
Allow for day of months scheduled alerts

Shiv-Chata:


src/components/ChatContent/ChatContent.js
```diff
@@ -364,20 +366,40 @@ export default class ChatContent extends React.Component {
     return this.state.isQueryRunning || this.state.isDrilldownRunning
   }
 
+  shouldHideQueryInputComponent = () => {
+    if (this.props.hideChatBarAfterInitialResponse && this.state.messages.length > 0) {
```
alexzhou-chata, line 370: const { hideChatBarAfterInitialResponse, shouldRender } = this.props
 const { messages } = this.state
 How about this?

Shiv-Chata: Updated

alexzhou-chata (approved):

PR #899, opened 2024-09-20
ANT-2965

eleon-chata-ai:
- update row count over 50k o 50000+ instead of 50001

### DIFF
--- a/src/components/ChataTable/ChataTable.js
+++ b/src/components/ChataTable/ChataTable.js
@@ -1378,7 +1378,7 @@ export default class ChataTable extends React.Component {
     return (
       <div className='table-row-count'>
         <span>{`Scrolled ${currentRowsFormatted} / ${
-          totalRowCount > 50000 ? rowLimitFormatted + '+' : totalRowsFormatted
+          totalRowCount > rowLimit ? rowLimitFormatted + '+' : totalRowsFormatted
         } rows`}</span>
       </div>
     )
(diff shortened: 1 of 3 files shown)


src/components/ChataTable/ChataTable.js
```diff
@@ -1374,10 +1372,14 @@ export default class ChataTable extends React.Component {
...
     const currentRowsFormatted = new Intl.NumberFormat(languageCode, {}).format(currentRowCount)
     const totalRowsFormatted = new Intl.NumberFormat(languageCode, {}).format(totalRowCount)
+    const rowLimit = this.props.response?.data?.data?.row_limit
+    const rowLimitFormatted = new Intl.NumberFormat(languageCode, {}).format(rowLimit)
 
     return (
       <div className='table-row-count'>
-        <span>{`Scrolled ${currentRowsFormatted} / ${totalRowsFormatted} rows`}</span>
+        <span>{`Scrolled ${currentRowsFormatted} / ${
+          totalRowCount > 50000 ? rowLimitFormatted + '+' : totalRowsFormatted
```
Shiv-Chata, line 1381: What does the network response when the data returned > 50000? I think there will be a key `count_sql` that will = 1. I think this would be a better condition to check

Shiv-Chata (commented):
Left a comment

src/components/ChataTable/ChataTable.js
eleon-chata-ai: resolved, used data set value vs hard coded value

Shiv-Chata (approved):

PR #900, opened 2024-09-25
Make Custom column name feature available in:

Shiv-Chata:
- DM component
- Dashboards
- Data Alerts


src/components/AddColumnBtn/CustomColumnModal.js
```diff
@@ -501,6 +509,26 @@ export default class CustomColumnModal extends React.Component {
     return !!this.state.selectedFnType && !!this.state.selectedFnColumn
   }
 
+  checkColumnName = (nameVal) => {
+    const columnNameExists = this.props?.columns?.find((col) => col?.display_name === nameVal)
+    if (columnNameExists || !nameVal) {
```
eleon-chata-ai, line 514: can remove if to return !(columnNameExists || !nameVal)

src/components/Input/Input.scss
```diff
@@ -36,6 +37,7 @@
       }
 
       &:hover {
+
```
eleon-chata-ai, line 40: extra white space

src/components/Input/Input.scss
```diff
@@ -276,4 +284,4 @@
     margin: 0 3px;
     opacity: 0.6;
   }
-}
+}
```
eleon-chata-ai, line 287: don't think the extra space at the bottom of the page is needed?

eleon-chata-ai (changes requested):

src/components/Input/Input.scss
Shiv-Chata: This is still at the end of the file, with no extra space. It is just that it has moved down the file from line 279 to 287

eleon-chata-ai (changes requested):

src/components/Input/Input.scss
Shiv-Chata: I think that fixed it

eleon-chata-ai (approved):

PR #908, opened 2024-10-03
ANT-2942

eleon-chata-ai:
- sorting-not-working-on-pivot-table

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
   },
   "dependencies": {
     "@react-icons/all-files": "^4.1.0",
+    "@types/luxon": "^3.4.2",
     "autoql-fe-utils": "1.6.4",
     "axios": "^1.4.0",
     "classnames": "^2.5.1",
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "lodash.has": "^4.5.2",
     "lodash.isempty": "^4.4.0",
     "lodash.isequal": "^4.5.0",
+    "luxon": "^3.5.0",
     "make-coverage-badge": "^1.2.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -80,6 +81,7 @@
     "lodash.has": "^4.5.2",
     "lodash.isempty": "^4.4.0",
     "lodash.isequal": "^4.5.0",
+    "luxon": "^3.5.0",
```
Shiv-Chata, line 84: What is this pkg being used for? Is it still being used or can it be removed?

src/components/ChataTable/TableWrapper.js
```diff
@@ -85,6 +85,7 @@ export default class TableWrapper extends React.Component {
   instantiateTabulator = () => {
     // Instantiate Tabulator when element is mounted
     this.tabulator = new Tabulator(this.tableRef, {
+      debugInvalidOptions: false,
```
Shiv-Chata, line 88: Is this the right solution to the problem of those console logs warnings showing up? Or should could we handle the invalid options better?

package.json
eleon-chata-ai: tabulator lib we are using for tables requires luxon and type/luxon for date sorting 
Dependency Required
You will need to include the luxon.js library to use this sorter

eleon-chata-ai (commented):
> Is this the right solution to the problem of those console logs warnings showing up? Or should could we handle the invalid options better?  

From reading online, this is what most have suggested as it was an issue in 5.4.4 Tabulator. But I'll pull the pr back and see what other options there are for both fixes

PR #909, opened 2024-10-03
ANT-2942

eleon-chata-ai:
- sorting not working on pivot table

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "ISC",
       "dependencies": {
         "@react-icons/all-files": "^4.1.0",
+        "@types/luxon": "^3.4.2",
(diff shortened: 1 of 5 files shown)


alexzhou-chata (commented):
you didn't update the version number

alexzhou-chata:
you have to run `npm install` in the root folder after you update the version, so that the package-lock.json will get updated with that version.

eleon-chata-ai (commented):
> you didn't update the version number 

updated

eleon-chata-ai (commented):
> you have to run `npm install` in the root folder after you update the version, so that the package-lock.json will get updated with that version. 

updated and pushed

alexzhou-chata (approved):

PR #973, opened 2025-01-30
ANT-3281

eleon-chata-ai:
sorting and filtering


alexzhou-chata (approved):

package.json
```diff
@@ -61,107 +61,107 @@
...
+    "react-toastify": "8.2.0",
+    "react-tooltip": "5.22.0",
+    "recordrtc": "5.6.1",
+    "sanitize-html": "2.3.2",
+    "sql-formatter": "2.3.3",
+    "tabulator-tables": "5.6.1",
     "uuid": "9.0.0"
   },
   "devDependencies": {
-    "@babel/cli": "^7.16.0",
```
Shiv-Chata, line 112: Thanks for making these exact!

src/components/ChataTable/ChataTable.js
```diff
@@ -63,6 +65,7 @@ export default class ChataTable extends React.Component {
     this.isFiltering = false
     this.isSorting = false
     this.pageSize = props.pageSize ?? 50
+    this.useRemote = this.props.response?.data?.data?.count_rows > 50000 ? 'remote' : 'local'
```
Shiv-Chata, line 68: Would be nice if 50000, 'remote' and 'local' were added to our constants list

src/components/QueryOutput/QueryOutput.js
```diff
@@ -925,7 +927,7 @@ export class QueryOutput extends React.Component {
           ...getAutoQLConfig(this.props.autoQLConfig),
           source: this.props.source,
           scope: this.props.scope,
-          debug: queryRequestData?.translation === 'include',
+          debug: queryRequestData?.translation,
```
Shiv-Chata, line 930: Has this endpoint been updated to accept a string instead of the boo?

Shiv-Chata (commented):
Looks good, just some questions and nice to haves. 

Did we implement something that would indicate to the user that the data they are looking at is a subset of the data when > 50000 and the sorting/filter won't be applied to all the data?

PR #1001, opened 2025-03-04
ANT-3414

eleon-chata-ai:
- Notification data filtering not working change duplicate column name

### DIFF
--- a/src/components/QueryOutput/QueryOutput.js
+++ b/src/components/QueryOutput/QueryOutput.js
@@ -1838,11 +1838,21 @@ export class QueryOutput extends React.Component {
       return null
     }
 
+    const usedNames = new Map()
     const formattedColumns = columns.map((col, i) => {
       const newCol = _cloneDeep(col)
 
       newCol.id = col.id ?? uuid()
       newCol.field = `${i}`
+      if (newCol.name) {
+        if (usedNames.has(newCol.name)) {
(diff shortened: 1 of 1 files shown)


src/components/QueryOutput/QueryOutput.js
```diff
@@ -1838,11 +1838,21 @@ export class QueryOutput extends React.Component {
...
     const formattedColumns = columns.map((col, i) => {
       const newCol = _cloneDeep(col)
 
       newCol.id = col.id ?? uuid()
       newCol.field = `${i}`
+      if (newCol.name) {
+        if (usedNames.has(newCol.name)) {
+          const count = usedNames.get(newCol.name) + 1
+          usedNames.set(newCol.name, count)
+          newCol.name = `${newCol.name} (${count})`
```
Shiv-Chata, line 1851: What happens when you send in the newCol to the BE? I can't imagine that is get's recognized

Shiv-Chata (commented):
Left a comment

src/components/QueryOutput/QueryOutput.js
eleon-chata-ai: would you have a query we can try that would have duplicate columns?

eleon-chata-ai: going to cancel the pr, this seems to be a problem. good catch

PR #1008, opened 2025-03-13
In PWA, remove the extra space on the notification page

alexzhou-chata:


src/js/DeviceInfoProvider.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { isMobile, isMobileLandscape } from './breakpoints'
+
+export default class DeviceInfoProvider extends React.Component {
```
Shiv-Chata, line 4: You could use newer react hooks like useEffect here. New components in react-autoql should as we try to move this package frwd. You could simply ask a AI tool to help with the conversion

Shiv-Chata (commented):
Left a comment

Shiv-Chata (approved):

PR #1031, opened 2025-04-07
ANT-3018-analytic-functions

eleon-chata-ai:


alexzhou-chata (approved):

example/src/App.js
```diff
@@ -943,6 +945,14 @@ export default class App extends Component {
           }}
           value={this.state.quantityDecimals}
         />
+        <h4>Number of Decimals for Ratio Values</h4>
+        <InputNumber
```
Shiv-Chata, line 949: Is the example app even being used anymore?

Shiv-Chata (commented):
Is this to address the date formatting issue we discussed yesterday? I don't see how the date precision is being used to determine the format to use.

PR #1050, opened 2025-04-25
Ant 3379 persist sort and filter data in dashboard tile 

Shiv-Chata:
- pass filter, and sort data from dashboard to query
- update OptionToolbar to include filter and sort data when creating dashboard tile from DM
- pass intialTableParams from dashboard tile all the way through the ChataTable component
- Update to version 8.10.0 and autoql-fe-utils

### DIFF
--- a/src/components/QueryOutput/QueryOutput.js
+++ b/src/components/QueryOutput/QueryOutput.js
@@ -65,6 +65,8 @@ import {
   getCleanColumnName,
   isDrilldown,
   CustomColumnTypes,
+  formatFiltersForTabulator,
+  formatSortersForTabulator,
 } from 'autoql-fe-utils'
 
 import { Icon } from '../Icon'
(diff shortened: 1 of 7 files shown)


src/components/QueryOutput/QueryOutput.js
```diff
@@ -1278,21 +1287,36 @@ export class QueryOutput extends React.Component {
...
+      const dataConfig = {
+        tableConfig: this.tableConfig,
+        pivotTableConfig: this.pivotTableConfig,
+      }
+      this.props.onColumnChange(
+        response?.data?.data?.fe_req?.display_overrides,
+        this.state.columns,
+        response?.data?.data?.fe_req?.additional_selects,
+        response,
+        dataConfig,
```
eleon-chata-ai, line 1306: we pass response here already, do we need all the other params coming from the response?

Shiv-Chata: You are correct, but it is a little interesting how the dashboardsTile component handles the parameters - yes a lot of them could be replaced by just passing the single response object through but then it would have to broken up in the receiving function anyways and i think we should wait to redo this piece once we have the time to fully update the whole component.

eleon-chata-ai (approved):

PR #1054, opened 2025-04-28
Added feature One click copy value in Data Messenger

alexzhou-chata:


eleon-chata-ai (dismissed):

src/components/QueryOutput/QueryOutput.js
```diff
@@ -1906,14 +1908,86 @@ export class QueryOutput extends React.Component {
...
+        valueContainer.className = 'react-autoql-cell-value'
+        const formattedValue = formatElement({
+          element: cellValue,
           column: newCol,
           config: getDataFormatting(this.props.dataFormatting),
           htmlElement: cell.getElement(),
         })
+        valueContainer.innerHTML = formattedValue
+        wrapper.appendChild(valueContainer)
+        if (cellValue !== null && cellValue !== undefined && cellValue !== '') {
```
Shiv-Chata, line 1934: if (cellValue) or if (!!cellValue) does not work here?

src/components/QueryOutput/QueryOutput.js
```diff
@@ -1906,14 +1908,86 @@ export class QueryOutput extends React.Component {
...
+            })
+          })
+        }
+
+        return wrapper
+
+        function copyToClipboard(text, element) {
+          const textarea = document.createElement('textarea')
+          textarea.value = text
+          textarea.style.position = 'absolute'
```
Shiv-Chata, line 1960: A className would be better no?

src/components/QueryOutput/QueryOutput.js
alexzhou-chata: If cellValue is 0 or false, these are valid values that should be displayed and made copyable, but they would be treated as falsy and would not pass the condition if (cellValue) or if (!!cellValue).

Shiv-Chata: Oh okay, you could simplify it a bit by changing it to 
if (cellValue != null && cellValue !== '') {}
as cellValue != null checks for both null and undefined

eleon-chata-ai: > from google ai - it is more compact. However, for clarity and to avoid potential confusion arising from loose comparison's type coercion, many developers prefer using strict comparison (!==) explicitly checking for both null and undefined.

Shiv-Chata (approved):

PR #1071, opened 2025-05-27
ANT-3281

eleon-chata-ai:
---client-side-sorting-and-filtering

### DIFF
--- a/src/components/ChatMessage/ChatMessage.js
+++ b/src/components/ChatMessage/ChatMessage.js
@@ -346,6 +346,7 @@ export default class ChatMessage extends React.Component {
             createDataAlertCallback={this.props.createDataAlertCallback}
             customOptions={this.props.customToolbarOptions}
             popoverAlign='end'
+            showFilterBadge={this.state?.localRTFilterResponse?.data?.data?.fe_req?.filters?.length > 0}
           />
         ) : null}
       </div>


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-autoql",
-  "version": "8.10.16",
+  "version": "8.10.17",
```
Shiv-Chata, line 3: The next version for his new feature should be v8.11.0 

https://semver.org/#:~:text=MINOR%20version%20when%20you%20add%20functionality%20in%20a%20backward%20compatible%20manner

src/components/ChataTable/ChataTable.js
```diff
@@ -103,13 +103,18 @@ export default class ChataTable extends React.Component {
...
       this.tableOptions.pagination = false
       this.tableOptions.paginationMode = this.useRemote
-      this.tableOptions.progressiveLoad = 'scroll' // v4: ajaxProgressiveLoad
-      this.tableOptions.ajaxURL = 'https://required-placeholder-url.com'
       this.tableOptions.paginationSize = this.pageSize
       this.tableOptions.paginationInitialPage = 1
-      this.tableOptions.ajaxRequesting = (url, params) => this.ajaxRequesting(props, params)
-      this.tableOptions.ajaxRequestFunc = (url, config, params) => this.ajaxRequestFunc(props, params)
-      this.tableOptions.ajaxResponse = (url, params, response) => this.ajaxResponseFunc(props, response)
+      if (this.useRemote === 'local') {
```
Shiv-Chata, line 108: Just a heads up, when using string values like 'local', we should check to see how many times this value is used and if it 3 or more we should create a constant string value for. I checked for 'local' and it looks like it is only being used three times with the updates to this PR. 

For reference: https://next.sonarqube.com/sonarqube/coding_rules?open=java%3AS1192&rule_key=java%3AS1192

src/components/ChataTable/ChataTable.js
```diff
@@ -103,13 +103,18 @@ export default class ChataTable extends React.Component {
...
+        this.tableOptions.progressiveLoad = 'load'
+        this.tableOptions.data = props.response?.data?.data?.rows
+      } else {
+        this.tableOptions.progressiveLoad = 'scroll' // v4: ajaxProgressiveLoad
```
Shiv-Chata, line 112: Is this comment here for a reason?

src/components/ChataTable/ChataTable.js
```diff
@@ -1401,18 +1403,42 @@ export default class ChataTable extends React.Component {
...
-      totalRowCount = this.props.data?.length
-    } else {
-      totalRowCount = this.props.response?.data?.data?.count_rows
+    if (this.useRemote === 'remote') {
+      if (this.props.pivot) {
+        totalRowCount = this.props.data?.length
+      } else {
+        totalRowCount = this.props.response?.data?.data?.count_rows
+      }
+    } else if (this.useRemote === 'local') {
```
Shiv-Chata, line 1423: Calculating the row count by comparing the size of the table does not seem like a stable and consistent solution. Have you considered and tried using 
```let rowCount = this.ref?.tabulator?.getDataCount('active')```? 

And if so, what issues did you run into?

File: src/components/ReverseTranslation/ReverseTranslationV2.js [hunk 5]
```diff
@@ -266,14 +270,28 @@ const ReverseTranslation = ({
...
+    const newParsedInterpretation = filterResponse?.data?.data?.parsed_interpretation
+    if (
+      initialParsedInterpretations?.current?.length &&
+      !deepEqual(newParsedInterpretation, initialParsedInterpretations?.current)
+    ) {
+      initialParsedInterpretations.current = newParsedInterpretation
+      const newArray = constructRTArray(newParsedInterpretation)
+      executePrerequisites(newArray)
+    }
+  }, [filterResponse?.data?.data?.parsed_interpretation])
```
Shiv-Chata, line 293: Instead of having a dedicated filterResponse object, could we not update the queryResponse object instead? This way when the queryResponse object gets updated the changes will cascade down to all dependant components and will save us from having to manage two objects that basically hold the same information.

Shiv-Chata (commented):
Left some comments

src/components/ChataTable/ChataTable.js
eleon-chata-ai: yes we did and it doesn't return what is expected. in this case it returns a count of all rows since all rows have been loaded already

src/components/ChataTable/ChataTable.js
eleon-chata-ai: that was there from before, Nikki added it I assume because the variable has changed,

src/components/ReverseTranslation/ReverseTranslationV2.js
Shiv-Chata: Right, this call only returns the RT but what i am suggesting is that instead of having a new FilterResponse object along side the original queryResponse is to update the RT and any other required fields in the queryResponse object from the result of this endpoint. 

i.e

RT call -> update RT in queryResponse object -> components update as before

src/components/ChataTable/ChataTable.js
Shiv-Chata: Do you mind showing me the issue sometime today? As i ran a quick test, filtering a date column under 50,000 rows and using getDataCount('active') i was able to get the correct number of rows. Maybe i am missing something tho

src/components/ReverseTranslation/ReverseTranslationV2.js
eleon-chata-ai: I'll need you to show me what you mean

Shiv-Chata (approved):

alexzhou-chata (approved):

PR #1072, opened 2025-06-03
ANT-3444 

eleon-chata-ai:
- add string conversion before replace


src/components/QueryOutput/QueryOutput.js
```diff
@@ -558,7 +558,7 @@ export class QueryOutput extends React.Component {
 
   hasError = (response) => {
     try {
-      const referenceIdNumber = Number(response.data.reference_id.split('.')[2])
+      const referenceIdNumber = Number(response.data?.reference_id?.split('.')[2])
```
Shiv-Chata, line 561: This logic works, but it's neither standard nor ideal.

Have you tried either of these options? 

`if (response.status >= 200 && response.status < 300) {
  return false;
}`

Or

`if (response.data?.message === "Success") {
  return true;
}`

src/components/QueryOutput/QueryOutput.js
```diff
@@ -1829,8 +1829,8 @@ export class QueryOutput extends React.Component {
           }
 
           // No logical operators detected, just compare numbers
-          const number = parseFloat(rowValue?.replace(/[^0-9.]/g, ''))
-          const filterNumber = parseFloat(headerValue?.replace(/[^0-9.]/g, ''))
+          const number = parseFloat(rowValue?.toString().replace(/[^0-9.]/g, ''))
+          const filterNumber = parseFloat(headerValue?.toString().replace(/[^0-9.]/g, ''))
```
Shiv-Chata, line 1833: This would also strip out any negative signs from the number? Can we guarantee that a negative value wouldn't be here?

Shiv-Chata (commented):
Left some comments

alexzhou-chata (approved):

alexzhou-chata (approved):

Shiv-Chata (approved):

PR #1073, opened 2025-06-11
ANT-3854 - Add vertical expand feature on Query Output in data messenger

alexzhou-chata:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-autoql",
-  "version": "8.11.1",
+  "version": "8.11.2",
```
Shiv-Chata, line 3: Since this is a new feature the version should be 8.12.0

For reference: https://semver.org/

src/components/ChatContent/ChatContent.js
```diff
@@ -396,6 +396,17 @@ export default class ChatContent extends React.Component {
     }
     return !shouldRender
   }
+  onMessageResize = (messageId) => {
+    if (this.messengerScrollComponent) {
```
Shiv-Chata, line 400: Couple things here: 

Instead of nesting everything inside an if, return early if the scroll component doesn’t exist:

```
if (!this.messengerScrollComponent) return;

  this.messengerScrollComponent.update();
  ...
}
```

You’re accessing this.state.messages multiple times, 

```
const { messages } = this.state;
const lastMessage = messages[messages.length - 1];
```

src/components/ChatMessage/ChatMessage.js
```diff
@@ -41,7 +43,19 @@ export default class ChatMessage extends React.Component {
       isSettingColumnVisibility: false,
       activeMenu: undefined,
       localRTFilterResponse: null,
+      isQueryOutputModalVisible: false,
+      // Add resize state
```
Shiv-Chata, line 47: Some of these comments from line 47 to 57 are not very useful or the variable name is descriptive enough. I would consider removing some of them.

src/components/ChatMessage/ChatMessage.js
```diff
@@ -416,6 +471,29 @@ export default class ChatMessage extends React.Component {
             </div>
           ) : null}
         </div>
+        {/* May use this later for resizable query output modal */}
```
Shiv-Chata, line 474: Did the product team not want this feature as well? 

Also, same as above

src/components/OptionsToolbar/OptionsToolbar.js
```diff
@@ -615,10 +631,12 @@ export class OptionsToolbar extends React.Component {
       <ErrorBoundary>
         <div
           className={`${isMobile ? 'react-autoql-toolbar-mobile' : 'react-autoql-toolbar'} options-toolbar
-            ${this.state.activeMenu ? 'active' : ''}
-            ${this.props.className || ''}`}
+			${this.state.activeMenu ? 'active' : ''}
+			${this.props.className || ''}`}
           data-test='autoql-options-toolbar'
         >
+          {/* Uncomment if you want to show the Open in New button */}
```
Shiv-Chata, line 638: I think we should remove all these comments if there is no immediate plan to enable this feature, as we want to maintain a production level code base

src/components/OptionsToolbar/OptionsToolbar.js
```diff
@@ -695,6 +713,8 @@ export class OptionsToolbar extends React.Component {
       const autoQLConfig = getAutoQLConfig(props.autoQLConfig)
 
       shouldShowButton = {
+        // uncomment this line if you want to show the Open in New button
```
Shiv-Chata, line 716: same as above

src/components/ChatMessage/ChatMessage.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { isMobile } from 'react-device-detect'
+// May use this later for resizable query output modal
```
Shiv-Chata, line 4: I think we should remove all these comments if there is no immediate plan to enable this feature, as we want to maintain a production level code base

Shiv-Chata (commented):
Left some comments

src/components/ChatMessage/ChatMessage.js
Shiv-Chata: Oh just saw your latest commit, looks good thanks!

Shiv-Chata (approved):
lgtm

PR #1076, opened 2025-06-23
ANT-3445 

eleon-chata-ai:
- fix badge when filtering on tile and notification

### DIFF
--- a/src/components/ChataTable/ChataTable.js
+++ b/src/components/ChataTable/ChataTable.js
@@ -69,7 +69,7 @@ export default class ChataTable extends React.Component {
     this.useRemote =
       this.props.response?.data?.data?.count_rows > TABULATOR_LOCAL_ROW_LIMIT
         ? LOCAL_OR_REMOTE.REMOTE
-        : LOCAL_OR_REMOTE.REMOTE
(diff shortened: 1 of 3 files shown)


alexzhou-chata (approved):

Shiv-Chata (dismissed):

Shiv-Chata (approved):
Why is this PR a draft PR?

eleon-chata-ai (commented):
> > Why is this PR a draft PR?  
> 
> I changed it Draft as I found there is another table, drillDownTable, that doesn't call the render the same way, and I need to fix this as well 

ok updated back to not draft

Shiv-Chata (approved):

Shiv-Chata:
> > > Why is this PR a draft PR?
> > 
> > 
> > I changed it Draft as I found there is another table, drillDownTable, that doesn't call the render the same way, and I need to fix this as well
> 
> ok updated back to not draft

Thanks. Ideally, we would like PR's to only be opened when the dev is done adding to it. I know this is a bit of different case but adding commits after a PR is open makes it more difficult to review as we don't know when it is ready.

PR #1088, opened 2025-07-22
AQLP-91---fix-date-range-picker-bug

eleon-chata-ai:
update function doing setting start end dates

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "ISC",
       "dependencies": {
         "@react-icons/all-files": "4.1.0",
-        "autoql-fe-utils": "1.7.9",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -4609,6 +4609,31 @@
         "source-map": "^0.5.7"
       }
     },
+    "node_modules/babel-plugin-emotion/node_modules/babel-plugin-macros": {
```
alexzhou-chata, line 4612: why we have those babel things here?

alexzhou-chata: still here?

alexzhou-chata (approved):
lgtm

Shiv-Chata (approved):

PR #1090, opened 2025-07-22
use fitColumns for non drilldown dashboard only

alexzhou-chata:


src/components/ChataTable/TableWrapper.js
```diff
@@ -26,7 +26,7 @@ export default class TableWrapper extends React.Component {
       reactiveData: false,
       autoResize: true,
       rowHeight: 25,
-      layout: this.props.pivot ? 'fitDataFill' : 'fitColumns',
+      layout: this.props.isDrilldown ? 'fitDataFill' : (this.props.scope === 'dashboards' ? 'fitColumns' : 'fitDataFill'),
```
Shiv-Chata, line 29: What was the issue with using fitColumns for drilldowns

alexzhou-chata: for drilldowns the table has lots of columns to view the data and it's better to use the fitDataFill, otherwise some numeric column's width will be minimal width automatically and affect the view.
Our initial purpose of this feature is to help the resize window scenario. 
Drilldown and other UI such as notification, data messenger, which are using tables are actually not being resized.

alexzhou-chata: Screenshot 2025-07-22 at 3 13 59 PM
but if we use `fitDataFill `, the above image demos that it has an expected UI.

Shiv-Chata: Okay, thanks for the explanation Alex. Sounds good

Shiv-Chata (approved):

PR #1091, opened 2025-07-23
Chart resizing performance

NikkiMeganMoore:
I removed the white overlay when resizing charts, and adjusted some of the chart layout refresh values so we can show the charts while they are being resized without flickering.

I also added a throttled refresh to the ChataChart component that will call continuously while the isResizing prop is true. This is used for when the DM is being resized or when a dashboard tile is being resized. 

The throttle value depends on how big the dataset is. It takes the number of data points and multiplies it by 2, with a min of 100ms and a max of 1000ms. For example: if a chart has 250 bars, the refresh rate will be 500ms.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "react-autoql",
-  "version": "8.12.13",
+  "version": "8.12.14",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "react-autoql",
-      "version": "8.12.13",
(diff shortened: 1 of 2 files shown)


Shiv-Chata:
Hey @NikkiMeganMoore all the additions and changes look good here, but you just need to update the pkg version

alexzhou-chata (changes requested):
we always update the version number and run `npm i` right now to keep the version.
Other things looks good to me.

PR #1096, opened 2025-07-30
AQLP-130: Dashboard toolbar widget

NikkiMeganMoore:
I put all dashboard tools into a react-autoql widget. This is to prepare for the eventual feature of dashboard slicing when we will want to add more complex tools to the toolbar. Then we can reuse these tools elsewhere outside of webapp.

### DIFF
--- a/src/components/DashboardToolbar/DashboardToolbar.js
+++ b/src/components/DashboardToolbar/DashboardToolbar.js
@@ -145,23 +145,25 @@ export class DashboardToolbarWithoutRef extends React.Component {
                   tooltipID={this.props.tooltipID}
                   onClick={this.props.onRefreshClick}
                 />
-                <Popover
-                  align='end'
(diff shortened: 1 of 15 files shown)


src/components/DashboardToolbar/DashboardToolbar.js
```diff
@@ -0,0 +1,266 @@
...
+                  tooltip='Close without saving'
+                  tooltipID={this.props.tooltipID}
+                  onClick={() => this.setState({ isConfirmCloseModalOpen: true })}
+                >
+                  Cancel
+                </Button>
+                <Button
+                  type='primary'
+                  icon='save'
+                  tooltip='Save and close'
```
eleon-chata-ai, line 223: capitalize close?

eleon-chata-ai (approved):

PR #1107, opened 2025-08-14
ANT-3859

eleon-chata-ai:
- sort and filter using useInfiniteScroll variable with RT updating and counts scrolling


src/components/ChataTable/ChataTable.js
```diff
@@ -478,20 +476,28 @@ export default class ChataTable extends React.Component {
...
         this.props.onSorterCallback(sorters)
       }
       this.setLoading(false)
     }
   }
 
-  onDataFiltering = () => {
+  onDataFiltering = (filters, rows) => {
     if (this._isMounted && this.state.tabulatorMounted) {
+      this.isFiltering = true // Set this to true when filtering starts
```
Shiv-Chata, line 488: Think this code is self explanatory and the comment is not really adding value

src/components/ChataTable/ChataTable.js
```diff
@@ -504,18 +510,33 @@ export default class ChataTable extends React.Component {
...
-        this.props.onFilterCallback(headerFilters, rows)
-      }
+      if (!this.useInfiniteScroll) {
+        const tableParamsFormatted = formatTableParams(
+          { ...this.tableParams, filter: headerFilters },
+          this.props.columns,
+        )
+
+        // Apply filters to original data to get true count
+        const filteredData = this.originalQueryData.filter((row) => {
```
Shiv-Chata, line 520: This is a pretty expensive bit of logic, i would consider moving it to its own function and updating it. 

Right now, the code applies each filter by looping over every row and calling filterDataByColumn on a single row array. This repeats work unnecessarily, for each row/filter pair we look up the column index with find and create a temporary array just to pass into filterDataByColumn.

Instead, we can build a Map of columnId → columnIndex once before, then iterate through filters and apply filterDataByColumn to the entire list each time using .reduce(). This reduces the filtering process a lot and makes the logic easier to read.

src/components/Dashboard/DashboardTile/DashboardTile.js
```diff
@@ -769,7 +783,7 @@ export class DashboardTile extends React.Component {
       queryResponse,
       dataConfig,
       displayOverrides,
-      tableFilters,
+      tableFilters, // This will update the filters
```
Shiv-Chata, line 786: Remove comment

src/utils/dataSourceUtils.js
```diff
@@ -0,0 +1,41 @@
+import _cloneDeep from 'lodash/cloneDeep'
```
Shiv-Chata, line 1: Is this file being used?

src/components/ChataTable/ChataTable.js
```diff
@@ -102,28 +104,24 @@ export default class ChataTable extends React.Component {
     }
 
     if (props.response?.data?.data?.rows?.length) {
-      this.tableOptions.sortMode = this.useRemote // v4: ajaxSorting = true
-      this.tableOptions.filterMode = this.useRemote // v4: ajaxFiltering = true
+      this.tableOptions.sortMode = 'remote' // v4: ajaxSorting = true
```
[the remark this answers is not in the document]
eleon-chata-ai: good catch, resolved

src/components/ChataTable/ChataTable.js
```diff
@@ -102,28 +104,24 @@ export default class ChataTable extends React.Component {
...
+      this.tableOptions.filterMode = 'remote' // v4: ajaxFiltering = true
```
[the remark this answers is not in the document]
eleon-chata-ai: good catch, resolved

src/components/ChataTable/ChataTable.js
```diff
@@ -394,9 +392,9 @@ export default class ChataTable extends React.Component {
       return
     }
 
-    this.ref.tabulator.options.sortMode = this.useRemote
-    this.ref.tabulator.options.filterMode = this.useRemote
-    this.ref.tabulator.options.paginationMode = this.useRemote
+    this.ref.tabulator.options.sortMode = 'remote'
+    this.ref.tabulator.options.filterMode = 'remote'
+    this.ref.tabulator.options.paginationMode = 'remote'
```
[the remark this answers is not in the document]
eleon-chata-ai: good catch, resolved

src/utils/dataSourceUtils.js
eleon-chata-ai: was at one point, removed

Shiv-Chata (commented):
Thanks for making those changes LGTM, but there are some merge conflicts and i would try to get another set of eyes on this to.

PR #1119, opened 2025-08-25
ANT-3859

eleon-chata-ai:
revert a few changes related to useInfiniteScroll branch

### DIFF
--- a/src/components/ChataTable/ChataTable.js
+++ b/src/components/ChataTable/ChataTable.js
@@ -1120,7 +1120,7 @@ export default class ChataTable extends React.Component {
         ?.getColumns()
         ?.find((col) => col.getField() === this.tableParams.filter[0]?.field)
 
-      column.scrollTo('middle')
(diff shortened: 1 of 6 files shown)


alexzhou-chata (approved):
LGTM

src/components/ChataTable/ChataTable.js
```diff
@@ -1447,7 +1447,7 @@ export default class ChataTable extends React.Component {
   }
 
   renderTableRowCount = () => {
-    if (this.isTableEmpty()) {
+    if (this.isTableEmpty() || !this.props.useInfiniteScroll) {
```
NikkiMeganMoore, line 1450: This returns null if infinite scroll is false? Dont we still want to include the row count in this case? For example, in all notifications

eleon-chata-ai: ok removed

NikkiMeganMoore (approved):

PR #1125, opened 2025-08-28
Table filter bugs

NikkiMeganMoore:
Fix inconsistent filtering in dashboards
- Added a callback for onTableParamsChange
- Called updateToolbars on filter changes
- Reverted showFilterBadge to use queryoutput ref again (it wont always be in sync the other way)


alexzhou-chata (approved):
LGTM

PR #1129, opened 2025-08-29
ANT-3859

eleon-chata-ai:
- use infinite scroll save and discard tile edits


NikkiMeganMoore (commented):
Can you review the current state of webapp staging? Some of these changes may not be necessary anymore if the table filters are performing as expected now

PR #1148, opened 2025-09-15
AQLP-42 

eleon-chata-ai:
- add sum at bottom of column for quanity and dollar_amt


alexzhou-chata (changes requested):
Recommendation: use function from utility library if we have it. If we don't have it, we can add it there.

PR #1153, opened 2025-09-19
AQLP-42 

eleon-chata-ai:
- fix pivot and add avg change from tabulator to react based


src/components/QueryOutput/QueryOutput.js
```diff
@@ -2185,6 +2186,7 @@ export class QueryOutput extends React.Component {
       document.body.removeChild(textarea)
     }
   }
+
```
alexzhou-chata, line 2189: do we need this empty row?

alexzhou-chata (approved):

PR #1381, opened 2026-06-02
AQLP 642 

eleon-chata-ai:
- reset query, parentheses fix that zach mentioned yesterday, memory optimizations


alexzhou-chata:
## Code Review — PR #1381 "AQLP 642"

**Scope:** (1) reset a single dashboard tile to default query state, (2) parentheses/custom-column fix, (3) memory optimizations. *(Test files excluded from this review.)*

### Overview
Adds an `enableResetQuery` feature to reset one tile (clearing filters/columns/sorts/aggregations/chart config while preserving query + title) with undo support that survives post-reset auto-saves; reworks `CustomColumnModal` parenthesis/division handling; and lands memory fixes (WeakMap listener dedup, undo-log caps, shallow tile cloning, unmount cleanup, a config-correction circuit breaker). The memory work and reset/undo bookkeeping are generally well-constructed. One real correctness regression, plus several behavior changes and dead-code items to confirm.

---

### 🔴 Must fix

**1. Broken regex-escape in `buildFnArray` silently breaks custom-column parsing** — `CustomColumnModal.js:561`

The character class was corrupted in this PR:
```js
// before (correct):
matchStr.trim().replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
// after (this PR):
const escaped = matchStr.trim().replace(/[.*+?^${}()|[\\]\\]/g, '\\$&')
```
The class now terminates early at the unescaped `]`, so it escapes **nothing** (verified empirically: `x(y)`→`x(y)`, `a.b`→`a.b`, `c[d]`→`c[d]`). `escaped` is then interpolated into `` new RegExp(`(^|...)(${escaped})(?=...)`) ``. For any column whose `name`/`table_column` contains regex metacharacters — essentially every SQL-expression column like `SUM(amount)`, `Profit (USD)`, `rate * 1.5` — placeholder substitution matches the wrong positions or (on unbalanced parens) throws and is swallowed by the `buildFnArray` try/catch (returns `[]`). Net effect: custom-column formulas referencing those columns silently fail to parse. Since this PR is partly about the custom-column/parenthesis path, I'd block on this.

**Fix:** restore `/[.*+?^${}()|[\]\\]/g`.

---

### 🟡 Confirm intent

**2. `isServerError` narrowed from all 5xx → only `.502`** — `DashboardTile.js:444` (used at 774)
Previously force-retried on any `5xx` status, `"internal server error"` message, or `*.500` reference; now only `*.502` (query-translation). Defensible, but a real change to error-recovery behavior — confirm 500-class failures are intentionally no longer auto-retried.

**3. `shouldComponentUpdate` lost its drag short-circuit** — `DashboardTile.js:256`
Removing `if (nextProps.isDragging && this.props.isDragging) return false` means every tile re-renders on each drag tick. May be deliberate (the new `setIsDragging` toggle drives child resizes), but it's a potential drag-perf regression on dashboards with many tiles. Worth a note in the PR.

---

### 🟠 Dead / orphaned in-repo code

These look like host-app integration hooks, but nothing in `src/` wires them up — please confirm external consumers exist, otherwise they're overhead/confusion:

- **`react-autoql:resetQuery` CustomEvent** (`OptionsToolbar.js:456`) is dispatched (with a non-trivial `queryInputState` capture) but **no listener exists anywhere in `src/`**. The reset already runs via the `onResetClick → resetTile` prop chain, so the event + capture currently do nothing in-repo.
- **`window.__qa_query_input_registry`** is only *read* in OptionsToolbar, never *written* in `src/`, so the `queryInputStateFromRegistry` branch is unreachable.
- **`QueryInput.restoreState`** (`QueryInput.js:95`) has no production caller (only its test). `getState` is used; `restoreState` isn't.
- **`config/node-crypto-polyfill.js`** is imported nowhere (rollup gets crypto via the new `--experimental-global-webcrypto` flag). The file appears dead — wire it into `rollup.config.js` or drop it and keep just the flag.

---

### ✅ Looks good
- Memory work is solid: WeakMap listener dedup that *replaces* rather than stacks (`ChataTable.js:1166`, `QueryOutput.js:2957`); `tileLog` capped at 20; `cloneTilesForLog` keeping responses shallow while deep-cloning only `fe_req`; `componentWillUnmount` clearing timers + nulling buffers; the `!this._isMounted` guard before `setParamsForTile`.
- `_consecutiveConfigResets` circuit breaker (`QueryOutput.js`) is a clean way to break the config-correction loop; gating `onTableConfigChange` on a valid config is the right call.
- `serialize-javascript: ^6.0.1` override is a worthwhile security pin.
- DataAlertModal fix (`DataAlertModal.js:230`) removes the `descriptionInput = currentDataAlert?.label` line that clobbered the description with the label object — genuine bug fix, plus a sensible `projectId` fallback and fetch-on-mount.
- Duplicate `componentDidMount` in Dashboard.js consolidated into one (the second class-field definition silently shadowed the first).

### Nits
- The `axisSorts` array→object normalization IIFE is duplicated verbatim for top/second halves (`DashboardTile.js:1946 & 2102`) — extract to a helper.
- `resetTile`'s catch path re-invokes `runSingleTile()`, which could double-execute on failure; low risk but worth a guard.

---

**Bottom line:** Fix the regex (#1) before merge — verified, silent regression in the custom-column path. #2–#3 just need a line of intent; the dead-code group should be wired up or removed. Test coverage looks extensive, which is reassuring given the intricacy of the reset/undo state machine.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

alexzhou-chata (approved):
LGTM

PR #1385, opened 2026-06-18
Simple table component

NikkiMeganMoore:
- Add SimpleTable component to eventually replace tabulator
- Convert most font sizes to rem instead of px
- Change styling for ChatContent to work in SI portal


src/components/ChatMessage/ChatMessage.js
```diff
@@ -754,6 +848,172 @@ export default class ChatMessage extends React.Component {
...
+          columns: queryResponse?.data?.data?.columns,
+          rows: filteredRows,
+        },
+        queryID: queryResponse?.data?.data?.query_id,
+        apiKey: auth.apiKey,
+        token: auth.token,
+        domain: auth.domain,
+      })
+
+      const { columns, rows } = response?.data ?? {}
```
alexzhou-chata, line 887: This destructures the result one level too shallow. The request payload above reads `columns` from `queryResponse?.data?.data?.columns`, the `catch` block reads `error?.response?.data?.data?.message`, and the sibling `fetchLLMSummary` call in this file consumes its result as `response?.data?.data?.summary`. Unless `fetchFollowOnQuery` unwraps differently from the other `fetch*` utils, `columns`/`rows` live at `response.data.data` — so as written they're both `undefined` and every follow-on result renders an empty `SimpleTable`.

```suggestion
      const { columns, rows } = response?.data?.data ?? {}
```

src/components/FollowOnModal/FollowOnModal.js
```diff
@@ -0,0 +1,204 @@
...
+          columns: queryResponse?.data?.data?.columns,
+          rows: filteredRows,
+        },
+        queryID: queryResponse?.data?.data?.query_id,
+        apiKey: auth.apiKey,
+        token: auth.token,
+        domain: auth.domain,
+      })
+
+      const { columns, rows } = response?.data ?? {}
```
alexzhou-chata, line 106: Same shallow-destructure as the inline follow-on thread in `ChatMessage`: the request reads `queryResponse?.data?.data?.columns` and the sibling `fetchLLMSummary` result is read at `response.data.data`, so the follow-on payload is almost certainly at `response.data.data` too. As written `columns`/`rows` are `undefined` and the result table is always empty.

```suggestion
      const { columns, rows } = response?.data?.data ?? {}
```

src/components/ChatMessage/ChatMessage.js
```diff
@@ -754,6 +848,172 @@ export default class ChatMessage extends React.Component {
...
+    this.setState({ followOnQueryText: e.target.value, followOnError: null })
+  }
+
+  handleFollowOnQueryKeyDown = (e) => {
+    if (e.key === 'Enter') {
+      this.handleFollowOnQuerySubmit()
+    }
+  }
+
+  handleFollowOnQuerySubmit = async () => {
```
alexzhou-chata, line 861: Two things vs. the equivalent `FollowOnModal.handleSubmit`:

1. **No early-out guard.** `handleFollowOnQueryKeyDown` calls this on every Enter regardless of input state, so an empty/whitespace question gets submitted (`question: ''`) and a second Enter re-fires while a request is still in flight. The submit button is `disabled`, but the keydown path bypasses it. `FollowOnModal.handleSubmit` guards with `if (!queryText.trim() || this.state.isLoading) return` — worth mirroring here.
2. **Unguarded setState after await.** The success and `catch` branches call `setState` without an `_isMounted` check (only `finally` has one), which can warn if the message unmounts mid-request.

src/components/ChatMessage/ChatMessage.js
```diff
@@ -969,6 +1235,15 @@ export default class ChatMessage extends React.Component {
 
   onDeleteMessage = () => this.props.deleteMessageCallback(this.props.id)
 
+  onFollowOnQueryResult = ({ columns, rows, question }) => {
```
alexzhou-chata: `onFollowOnQueryResult` is never referenced — follow-on results are rendered inline by `renderFollowOnQueryThread`. This handler, and the `FollowOnResult` import on line 32 that is only used here, look like dead code from an earlier approach. Safe to drop both.

src/components/Notifications/DataAlertsTabbed/DataAlertRow.js
```diff
@@ -0,0 +1,417 @@
...
+    if (this.props.dataAlert?.title && this.props.dataAlert.title !== prevProps.dataAlert?.title) {
+      this.setState({ title: this.props.dataAlert.title })
+    }
+  }
+
+  hasError = () =>
+    ['GENERAL_ERROR', 'EVALUATION_ERROR', 'UNRECOVERABLE'].includes(this.state.status)
+
+  isEnabled = () => DATA_ALERT_ENABLED_STATUSES.includes(this.state.status)
+  isDisabled = () => DATA_ALERT_STATUSES.UNRECOVERABLE.includes(this.state.status)
```
alexzhou-chata: `DATA_ALERT_STATUSES.UNRECOVERABLE` is the string `"UNRECOVERABLE"`, so this evaluates `"UNRECOVERABLE".includes(this.state.status)` — a substring test on the constant rather than a status comparison. It returns `true` for the exact `'UNRECOVERABLE'` status, but also returns `true` when `status` is `''` (every string contains the empty string) and is backwards in intent. Unlike `isEnabled` above (where `DATA_ALERT_ENABLED_STATUSES` really is an array), this should be equality:

```suggestion
  isDisabled = () => this.state.status === DATA_ALERT_STATUSES.UNRECOVERABLE
```

src/components/SimpleTable/SimpleTable.js
```diff
@@ -0,0 +1,645 @@
...
+      <tbody>
+        {topPad > 0 && (
+          <tr style={{ height: topPad }}>
+            <td colSpan={columns.length} style={{ padding: 0, border: 0 }} />
+          </tr>
+        )}
+        {visibleRows.slice(startIdx, endIdx).map((row, i) => {
+          const rowIndex = startIdx + i
+          return (
+            <tr key={rowIndex} className={`simple-table-row${rowIndex % 2 === 1 ? ' simple-table-row-even' : ''}`}>
```
alexzhou-chata: Nit: `simple-table-row-even` is applied when `rowIndex % 2 === 1`, i.e. to odd indices (the 2nd, 4th… rows). Striping renders fine, but the class name is inverted relative to the index it keys off of — consider `rowIndex % 2 === 0` or renaming the class to match.

alexzhou-chata (commented):
## Review: Simple table component

Large PR (113 files): a new virtualized `SimpleTable`, a tabbed Data Alerts UI, an inline + modal follow-on-query feature, the Sankey diagram re-enabled, and a broad px to rem font migration. The structure is clean and the mechanical work is sound — I spot-checked the rem conversions and they're all correct `/16` multiples, and the `ChataHistogram` null-filtering fix looks right.

Inline comments below. The ones worth resolving before merge:

### Functional bugs
- **Follow-on result tables render empty.** Both `ChatMessage` and `FollowOnModal` destructure `fetchFollowOnQuery`'s result as `response.data`, but the payload is at `response.data.data` — matching the request payload, the error handler, and the sibling `fetchLLMSummary` usage in the same file. See ChatMessage.js:885 and FollowOnModal.js:104.
- **Inline follow-on input has no submit guard.** Enter submits empty/whitespace questions and re-fires while a request is in flight; `FollowOnModal.handleSubmit` already guards this. ChatMessage.js:861.

### Smaller correctness / cleanup
- `DataAlertRow.isDisabled()` calls `.includes()` on a string constant instead of `===`. DataAlertRow.js:74.
- Dead code: `ChatMessage.onFollowOnQueryResult` and the `FollowOnResult` import that only it uses.
- Cosmetic: the `simple-table-row-even` class is applied to odd row indices.

### Note on SimpleTable as a tabulator replacement
The description says SimpleTable will "eventually replace tabulator," but `QueryOutput` still renders `ChataTable` on this branch (the initial swap was reverted), so SimpleTable is currently only used for read-only follow-on result display. Before it can replace `ChataTable` in `QueryOutput` it will need the ref methods `QueryOutput` calls — `setPageLoading`, `copyToClipboard`, `getBase64Data`, `addCustomColumn` — plus header-filter restore. Flagging for the radar; not a blocker for this PR's actual usage.

Nice touches: the `key={\`${type}-${dataAlert.id}\`}` collision fix in DataAlerts, the explicit `fill` attributes for Sankey PNG export, and the new loading-skeleton states.

src/components/ChatMessage/ChatMessage.js
NikkiMeganMoore: This is actually how the new response is from the magic wand service, it is correct the way it is

src/components/FollowOnModal/FollowOnModal.js
NikkiMeganMoore: Same as above

src/components/ChatMessage/ChatMessage.js
alexzhou-chata: Retracting this — it's **not** a bug, and you were right to leave it. I checked the published `autoql-fe-utils@1.11.23`: `fetchFollowOnQuery` resolves with `response.data` (`.then((response) => Promise.resolve(response.data))` — it unwraps the axios layer internally), unlike `fetchLLMSummary` which returns the full `response`. So `response.data` here is already the API payload and `response.data.columns/rows` is correct. My comparison to `fetchLLMSummary` was the wrong reference. Sorry for the noise.

src/components/FollowOnModal/FollowOnModal.js
alexzhou-chata: Same retraction as the ChatMessage thread — not a bug. `fetchFollowOnQuery` in `autoql-fe-utils@1.11.23` returns `response.data` (unwrapped one level), so `const { columns, rows } = response?.data` correctly reads the payload. Leaving the line as-is is right.

src/components/FollowOnModal/FollowOnModal.js
```diff
@@ -0,0 +1,207 @@
...
+      const question = queryText.trim()
+
+      this.setState(
+        (prev) => {
+          const results = [...prev.results, { id: Date.now(), question, columns, rows }]
+          this.props.onResultsChange?.(results)
```
alexzhou-chata: Minor: calling `this.props.onResultsChange?.(results)` here runs a side effect (a parent `setState` in `DashboardTile`) from inside this component's `setState` updater. Updaters are expected to be pure — under React 18 StrictMode they're invoked twice, so this would fire twice (harmless here since it's idempotent). Cleaner to move it to the updater's callback arg:

```js
this.setState(
  (prev) => ({ queryText: '', results: [...prev.results, { id: Date.now(), question, columns, rows }] }),
  () => {
    this.props.onResultsChange?.(this.state.results)
    setTimeout(() => this.scrollToBottom(), 50)
  },
)
```

Not a blocker.

alexzhou-chata (commented):
## Recheck — latest commits (48a5bc1)

Thanks for the updates. Walking the changes since my last pass:

### Addressed correctly
- ✅ Inline follow-on submit now guards empty/whitespace + in-flight, and the post-`await` `setState`s are `_isMounted`-guarded (ChatMessage).
- ✅ Dead `onFollowOnQueryResult` handler removed.
- ✅ `DataAlertRow.isDisabled()` now uses `===`.
- ✅ `simple-table-row-even` striping index fixed.

### Retracting my earlier call (the two were correct as written)
The two `response.data` vs `response.data.data` comments were my mistake — you were right to leave them. In the published `autoql-fe-utils@1.11.23`, `fetchFollowOnQuery` resolves with `response.data` (it unwraps the axios layer internally), whereas `fetchLLMSummary` returns the full `response` — so the conventions differ and `const { columns, rows } = response?.data` is correct. I've replied in both threads.

### New this round (persist follow-on threads in dashboards)
Looks good overall. Two small things:
- The dead-code removal orphaned `src/components/FollowOnResult/` (`FollowOnResult.js` + `.scss`) — nothing imports it anymore. Safe to delete unless it's intended for a later path.
- One inline note on the `onResultsChange` call placement.

Net: the substantive items are resolved. Remaining notes are minor / optional.

alexzhou-chata:
Recheck on `3acf0f6` — the `onResultsChange` call is now in the `setState` callback with a pure updater, which is exactly right. That clears the last note from my side; all substantive items are resolved. 👍

Only leftover is optional cleanup: `src/components/FollowOnResult/` (`.js` + `.scss`) is now orphaned — no importers — so it can be deleted whenever convenient.

alexzhou-chata (dismissed):

alexzhou-chata:
Recheck on `41b91d9` — orphaned `FollowOnResult` files removed cleanly; no dangling imports or exports left referencing them, and the follow-on UI uses `SimpleTable` directly. Everything from my review is resolved — nothing outstanding on my side. 🚀

alexzhou-chata (approved):

PR #1386, opened 2026-06-19
AQLP 792 

eleon-chata-ai:
- on DB save send query IDs to dashboard tiles api


alexzhou-chata:
## Code Review — AQLP-792 (capture & persist query IDs)

Thorough, well-tested PR. Roughly ~2000 of the +2816 lines are tests, and coverage of the new state machine is excellent. Nothing below is a hard blocker, but a few items are worth confirming before merge.

### Strengths
- **Test coverage is the standout.** Nearly every new method has direct unit tests covering guards, NaN/empty edges, and view-vs-edit branches (`getDirtyTileKeys`, `getFailedTiles`, `disableBaseFilterInputs`, `shouldShowDirtyBadge`, `areTopAndBottomSameQuery`, queryId capture w/ `isCachedRefresh`).
- Good defensive habits: `baseFilters` is `_cloneDeep`'d (with a test), and `sessionFilters` is now copied (`[...baseFilters]`) to avoid in-place mutation.
- Clean refactor extracting the duplicated axis-sorts IIFE into `normalizeAxisSorts`.
- Dirty/failed badge mutual-exclusion is consistent across `className`, `shouldShowDirtyBadge`, and tests.

### Correctness concerns

**1. `getFailedTiles` and `hasError` disagree on a missing `reference_id` (medium).**
`Dashboard.getFailedTiles` computes `Number(String(reference_id||'').split('.')[2])` → `NaN` when absent, and `NaN < 200 || NaN >= 300` is **false** → *not failed*. But `DashboardTile.hasError` returns **true** (error) for the same input. So a tile with a `queryId`+`queryResponse` but no `reference_id` is treated as errored by the toolbar (`hideReportProblem`, no `showRefreshInEdit`) yet shows **no red failed badge**. Consider one shared helper so "failed" and "error" can't diverge.

**2. Dirty gate can permanently block Save on suggestion/replacement responses (medium — confirm intended).**
`getDirtyTileKeys` marks any tile with `replacements`/`items` as dirty, and `hasDirtyTiles()` **disables the Save button**. If a tile legitimately resolves to a safety-net/suggestion response, the user may never be able to re-execute it into a non-suggestion state — locking them out of saving the whole dashboard. Is there an escape hatch (delete/revert the tile)? Worth confirming this can't trap users.

**3. View-mode behavior change: filters no longer force REMOTE (medium — intentional but broad).**
The old `useRemote` forced REMOTE when `fe_req.filters` or `initialTableParams.filter` were present; the new logic drops that outside edit mode, so a *filtered* view-mode table under `TABULATOR_LOCAL_ROW_LIMIT` is now LOCAL. The rationale (fe_req filters baked into cached SQL) is sound, but this affects **all** `ChataTable` consumers in view mode (DataMessenger, drilldowns, notifications), not just dashboards. Please confirm filtered client-side tables still paginate/sort correctly in those cases.

**4. `onUpdateFilterResponse` now gated behind `showQueryInterpretation` (low — confirm).**
The debounced `getRTForRemoteFilterAndSort` (reverse-translation refresh → `onUpdateFilterResponse`) now also requires `showQueryInterpretation`. Logically correct, but it's a behavior narrowing for any consumer relying on that callback firing without the flag.

**5. `disableBaseFilterInputs` is imperative and asymmetric (low).**
It mutates the DOM directly via `setTimeout(…, 0)` (guarded by `_isMounted`). I don't see an explicit **re-enable** path when transitioning back into edit mode — it appears to rely on a remount/`setFilters` recreating the inputs. Worth a manual check that toggling edit→view→edit doesn't leave base-filter inputs stuck disabled.

### Minor / polish
- **Unused destructured vars** in `Dashboard.addTile` (`const { queryResponse, secondQueryResponse, ...rest }`) — fine functionally but likely trips `no-unused-vars`.
- **Hard-coded hex + `!important`** in `DashboardTile.scss` (`#FFA700`, `#ff0000`, `border-top … !important`). Most of the library themes via SCSS variables — consider theme vars so badges respect dark mode.
- **Removed reset-fallback retry** in `Dashboard.js` (catch no longer re-runs `runSingleTile()`) — looks like deliberate simplification; confirm a failed reset still leaves the tile recoverable.
- **Public-ish API removal**: `QueryInput.getState`/`restoreState` + the `react-autoql:resetQuery` window event are deleted together. They look QA-internal (`window.__qa_query_input_registry`), but since this is a published library, double-check no consumer listens for that event.

### Security
No concerns. No new auth/network surface; query IDs are already part of the response contract. The deleted `config/node-crypto-polyfill.js` is **not referenced** anywhere (not in jest `setupFiles`, babel, or rollup), so removing it is safe dead-code cleanup. Dependency bumps move toward patched versions.

### Verdict
Solid, well-tested feature. I'd want answers on **#1** and **#2** before merge, plus a quick manual confirmation of **#3** given its blast radius beyond dashboards.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

eleon-chata-ai:
Fixed the getFailedTiles/hasError mismatch so missing reference_id values are consistently treated as failures. Also added an enableBaseFilterInputs recovery path in ChataTable to prevent inputs from getting stuck during edit → view → edit transitions, cleaned up unused destructured vars in addTile, and replaced hardcoded badge/border hex values with CSS variables.

The remaining review items are intentional: the dirty-gate behavior, the onUpdateFilterResponse / showQueryInterpretation gating, and the REMOTE → LOCAL view-mode filter change (limited to dashboard tiles and does not affect other consumers).

The QueryInput.getState / restoreState and react-autoql:resetQuery event concerns are false positives. These APIs were introduced and removed entirely within this branch, never shipped in a published package (latest npm release is 8.29.6 and they do not appear in src/index.d.ts). Searches across react-autoql, autoql-webapp-frontend, and caas-integration-portal found no external usage, so no deprecation notice or semver bump is required.

eleon-chata-ai:
banner matrix:

 Dashboard   │ Tile            │ Scenario                    │ isDirty │ isFailed │ save disabled
─────────────┼─────────────────┼─────────────────────────────┼─────────┼──────────┼───────────────
 New         │ New             │ Success run                 │  false  │  false   │      no
 New         │ New             │ DB/server error             │  false  │  true    │     YES
 New         │ New             │ Disambiguation (items)      │  false  │  true    │     YES
─────────────┼─────────────────┼─────────────────────────────┼─────────┼──────────┼───────────────
 Existing    │ New (added)     │ Success run                 │  false  │  false   │      no
 Existing    │ New (added)     │ DB/server error (bug fixed) │  false  │  true    │     YES
─────────────┼─────────────────┼─────────────────────────────┼─────────┼──────────┼───────────────
 Existing    │ Existing        │ Query unchanged             │  false  │  false   │      no
 Existing    │ Existing        │ Query changed, not re-run   │  true   │  false   │     YES
 Existing    │ Existing        │ Re-run → success            │  false  │  false   │      no
 Existing    │ Existing        │ DB/server error             │  false  │  true    │     YES
 Existing    │ Existing        │ Timeout (422)               │  false  │  true    │     YES
 Existing    │ Existing        │ Disambiguation (items)      │  true   │  true    │     YES
 Existing    │ Existing        │ Replacements                │  true   │  false   │     YES

alexzhou-chata:
Thanks for the quick turnaround — re-reviewed the latest push. 👍

**Confirmed resolved:**
- **#1 `getFailedTiles`/`hasError`** — now consistent (`!(refId >= 200 && refId < 300)`), and dropping the `!tile.queryId` gate so new tiles with errors/`items` are flagged is a nice catch beyond the original ask.
- **#5 base-filter inputs** — `enableBaseFilterInputs` recovery path + the edit→view→edit coverage addresses the asymmetry.
- Minor cleanups (unused destructure, themed CSS vars) all landed.

**Agreed / accepting:**
- The `getState`/`restoreState` + `react-autoql:resetQuery` removal — solid rebuttal, that's a false positive. Introduced and removed within this branch, never in a published release, no cross-repo usage. No action needed.
- The dirty/Save gate (#2) and `showQueryInterpretation` gating (#4) as intentional — the banner matrix is a great reference for the intended states.

**One small clarification on #3 (not a blocker):** the "limited to dashboard tiles" framing isn't quite what the code does — `useRemote` is computed identically for every `ChataTable` in the constructor, and the removed `fe_req.filters` / `initialTableParams.filter` branches weren't dashboard-gated, so filtered sub-limit tables in DataMessenger/drilldowns/notifications also shift to LOCAL. I think the **conclusion** (no harmful effect on other consumers) still holds — sub-limit results are fully loaded client-side, and query-level `fe_req` filters aren't clearable via header inputs — so this is just a note to confirm that reasoning rather than the dashboard-scoping. If a quick smoke test of a filtered view-mode DataMessenger result looks right, I'm happy.

Nice work on the test coverage throughout.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alexzhou-chata (approved):

PR #1388, opened 2026-06-26
AQLP 642   reset single tile

eleon-chata-ai:
Restores and persists dashboard tile filter state across remounts, including locked table filters, bottom-tile tracking, drilldown filter badges, and preventing drilldown params from leaking back to source tiles. Also replaces the TableWrapper 500ms timeout with a double-rAF layout sync and cleans up several prop/guard issues.

Fixes a bug where resetting a single tile after changing filters/sort in view mode and entering edit mode could carry stale view-mode state into edit mode. Also resolves several Tabulator race conditions and unhandled promise rejections triggered during rapid reset/reload flows.


src/components/QueryOutput/QueryOutput.js
```diff
@@ -3055,9 +3083,25 @@ export class QueryOutput extends React.Component {
       return newCol
     })
 
+    formattedColumns = this.reorderColumnsByUserOrder(formattedColumns)
```
alexzhou-chata: **High — column reorder corrupts the column↔row-data mapping**

`formatColumnsForTable` derives each column's Tabulator `field` positionally (`newCol.field = \`${i}\``, line 2936) while row arrays stay in backend order, so `columns[i].field === i` is an invariant the rest of this file relies on. `reorderColumnsByUserOrder` permutes that array *after* the fields are assigned, breaking it two ways:

1. **Index-based consumers misalign immediately.** `setTableConfig` → `getStringColumnIndices` / `getNumberColumnIndices` / `getDateColumnIndex` return *array positions*, and `processDrilldown` / `getFilterDrilldown` / `generateDatePivotData` then index the still-data-ordered rows with those positions. Drag a numeric column to position 0 → `numberColumnIndex` becomes 0 → charts/drilldowns read the string column's data.
2. **The corruption compounds across rebuilds.** `updateColumns` writes the reordered array into both `state.columns` and `this.queryResponse.data.data.columns` (line ~1304). The next rebuild (line ~564 re-reads `queryResponse.data.data.columns`; `_handleHiddenColumn` re-feeds `state.columns`) sends the already-reordered array back through `formatColumnsForTable`, which re-derives `field` from the *new* positions — so a column now points at a different row index and cells render under the wrong headers. In edit mode this mis-fielded array is also persisted via `onColumnChange`.

`movableColumns` is always enabled, so this is reachable in normal use.

**Fix:** keep `state.columns` (and the `tableConfig` indices) in canonical data order and apply the user's drag-order only at the Tabulator display layer; or make `field` idempotent by deriving it from a stable per-column identity (persistent key / original index) rather than the current array position.

src/components/QueryOutput/QueryOutput.js
```diff
@@ -617,8 +622,17 @@ export class QueryOutput extends React.Component {
...
       }
 
+      if (this.props.lockedFilters !== prevProps.lockedFilters || this.state.columns !== prevState.columns) {
+        this.formattedLockedFilters = this.formatLockedFilters(this.props.lockedFilters, this.state.columns)
+      }
+
       if (prevProps.isEditing !== this.props.isEditing) {
         if (this.props.isEditing) {
+          // Reset filters/sorters (otherwise only set on mount) so stale view-mode header filters don't leak into edit mode.
+          this.formattedTableParams = {
```
alexzhou-chata, line 656: **Medium — this edit-mode reset never runs for dashboard tiles**

This block is gated on `prevProps.isEditing !== this.props.isEditing` (line 629) → `this.props.isEditing`, but `DashboardTile` passes dashboard-edit state as `isDashboardEditing` (DashboardTile.js:1542) and never `isEditing`. So for a tile's `QueryOutput`, `this.props.isEditing` stays at its default `false`, and this reset (plus the pre-existing legend-state restore in the same block) is dead for dashboard tiles.

Concrete leak it was meant to prevent: in view mode the user sorts a column → `onTableParamsChange` updates `this.formattedTableParams.sorters`. DashboardTile's `viewModeFiltersDiffer` only compares `filters` (DashboardTile.js:1017), so no remount/refetch is forced on entering edit mode, and because this reset doesn't fire the stale view-mode sort survives — the next edit-mode `queryFn` that omits explicit `orders` re-applies it (`orders: this.formattedTableParams?.sorters`).

**Fix:** key this on `isDashboardEditing` (as `onNewData` already does at line 2183): `if (prevProps.isDashboardEditing !== this.props.isDashboardEditing) { if (this.props.isDashboardEditing) { …reset… } }`, moving the legend-state restore with it.

src/components/Dashboard/DashboardTile/DashboardTile.js
```diff
@@ -248,12 +257,33 @@ export class DashboardTile extends React.Component {
...
     const nextQRId = nextQR?.data?.data?.query_id
-    if (nextQR && !this.state.isTopExecuting && (prevQR === null || (prevQR && prevQRId !== nextQRId))) {
-      this.wasFilteringBeforeRemount = this.state.responseRef?.isFilteringTable() || false
+    // Self-reported query_id changes (onColumnChange) must not remount, or in-memory column state (frozen/drag-order) is lost
+    const isSelfColumnChange = !!nextQRId && nextQRId === this._columnChangeQueryId
+    if (
+      nextQR &&
+      nextQR !== prevQR &&
+      !this.state.isTopExecuting &&
+      (prevQR === null || (prevQR && prevQRId !== nextQRId && !isSelfColumnChange) || this._forceRemountOnNextResponse)
```
alexzhou-chata: **Medium — `_forceRemountOnNextResponse` can't fire the intended remount, then lingers and defeats the `isSelfColumnChange` guard**

The flag is set only at line 269 (view→edit when `viewModeFiltersDiffer`) and consumed only at line 286, inside this branch. But the edit-entry `processTile()` sets `isTopExecuting: true` (via `processTileTop`), cleared only later by `setTopExecuted` on Dashboard's `(i+1)*50ms` `debouncedOnChange` timer — *after* the fresh `queryResponse` has already propagated down. So:
- On the `componentDidUpdate` where `props.tile.queryResponse` changes, `isTopExecuting` is still true → the line 283 gate skips the bump.
- On the later update where `setTopExecuted` flips it false, `nextQR === prevQR` → the line 282 gate skips it.

Net: the intended remount never happens and the flag is never cleared. It then survives into a later, unrelated response: apply a header filter in view mode → enter edit mode (flag set) → add/edit a custom column. `onColumnChange` sets `_columnChangeQueryId` so `isSelfColumnChange` is true (the guard added right above to *preserve* frozen/reordered columns), but `|| this._forceRemountOnNextResponse` here still forces the `queryResponseVersion` bump, remounting `QueryOutput` and wiping exactly the column state the guard protects.

**Fix:** clear the flag when the edit-entry refetch settles (`.finally` on the `processTile()` at line 271) or when `isTopExecuting` transitions to false; and if a remount is genuinely needed on edit entry, bump `queryResponseVersion` directly after the refetch resolves instead of routing it through this `isTopExecuting`-gated path.

src/components/Dashboard/DashboardTile/DashboardTile.scss
```diff
@@ -710,6 +710,16 @@
   box-shadow: inset 0 3px 0 var(--react-autoql-warning-color, #ffa700), rgba(0, 0, 0, 0.1) 0px 4px 6px -1px, rgba(0, 0, 0, 0.06) 0px 2px 4px -1px !important;
 }
 
+.dashboard-tile-split-pane-container.top-dirty,
```
alexzhou-chata: **Low — dead split-pane CSS / scaffolding**

These `.dashboard-tile-split-pane-container.{top,bottom}-{dirty,failed}` rules (713–721) and `&.split-pane-badge` (744–747) target classes that aren't emitted anywhere in `src/` — grep finds them only in this file. The `getPaneBannerType()` helper (DashboardTile.js:430) that would supply the `dirty`/`failed` suffix is defined and unit-tested but never called in render, and `renderContent()` renders only the top response (there is no split-pane container element). This is unwired scaffolding — either wire `getPaneBannerType` into a split-view render, or drop these rules (and the related `splitView` guards in DashboardTile.js) until that rendering lands.

alexzhou-chata (commented):
## Automated review — PR #1388 (AQLP-642: reset single tile)

Reviewed the full diff (36 files, +3178/−705) via a multi-agent pass with independent adversarial verification of every candidate finding. Overall this is a large, mostly well-reasoned change: dashboard tile filter/sort/freeze/column-order persistence across remounts, locked-filter + drilldown-badge support, the `TableWrapper` 500 ms → double-`requestAnimationFrame` redraw swap, Tabulator race-condition / promise-rejection hardening, `MEDIAN` support + a Prettier/JSX cleanup in `CustomColumnModal`, and project-vs-management scoping for data-alert labels.

### Strengths
- The Tabulator hardening is careful: `?.catch?.()` guards, `blockRedraw`/`restoreRedraw` symmetry, `setSort` batching (fixes the old per-sorter loop that only kept the *last* sorter), and the `filterModule.tableInitialized` toggle to collapse N header-filter reloads into one.
- The double-rAF redraw is correctly guarded against unmount (`this._isMounted && this.tabulator`).
- Clean pure-helper extraction in `Dashboard.js` (`isResponseFailed` / `isTileFailed` / `isTileDirty`) with matching unit tests, plus strong new test coverage across the changed components.

### Findings (inline)
- **1 high** — I'd treat it as a blocker for the column-reorder feature.
- **2 medium** — correctness gaps in the exact view↔edit state-persistence paths this PR targets (both are "the code that was meant to prevent a leak never actually runs").
- **1 low** — dead CSS/scaffolding.

### Considered and dropped (~13 lower-confidence findings)
After adversarial verification I discarded a number of candidates, e.g.: the `DataAlertModal`/`CustomFilteredAlertModal` endpoint switch is benign because `getAllDataAlertsLabels` and `getAllDataAlertsLabelsByProject` are two API-tier views of the same label resource; the `redraw` rAF is already unmount-guarded; `clearSort()` has no stale-column-ref bug (unlike `clearHeaderFilter()`, which does); `formattedLockedFilters` one-render staleness is irrelevant since `ChataTable` only reads `lockedFilters` in its constructor.

### One thing worth confirming (not flagged inline)
`DataAlertModal` / `CustomFilteredAlertModal` now default `isManagementPortal={false}`, and the `DataAlerts` / `DataAlertsTabbed` wrappers don't forward it — so any existing **management-portal** consumer silently switches the *label fetch* from the management endpoint to the project endpoint. Verification indicates these are equivalent views of the same resource (so this is likely harmless), but if a management portal can surface a different label set than a single project, this is a behavior change worth a conscious decision.

<sub>🤖 Automated review generated with [Claude Code](https://claude.com/claude-code). Please sanity-check before acting.</sub>

eleon-chata-ai:
Confirmed against the review:

High (reorderColumnsByUserOrder corruption) — fixed: state.columns now stays canonical (never reordered); a new getOrderedTableColumns() applies drag-order only for the ChataTable display prop. Covered by the updated formatColumnsForTable test and the new getOrderedTableColumns test, both passing.
Medium (isEditing vs isDashboardEditing) — fixed: reset now checks prevProps.isEditing || prevProps.isDashboardEditing on both sides.
Medium (_forceRemountOnNextResponse race) — fixed: the flag is now consumed/cleared inside setTopExecuted after the refetch settles, instead of the isTopExecuting-gated branch that could never fire it.
Low (dead split-pane CSS) — fixed: the .top-dirty/.bottom-dirty/.top-failed/.bottom-failed/.split-pane-badge rules are removed.
No regressions in the full DashboardTile/QueryOutput test run. The isManagementPortal note wasn't part of these 4 changed files, so it's unaffected either way.

-------------------------

1. High — column reorder corruption

Run a query that returns a mix of string + number columns (e.g. a list query).
Switch to a chart, note which column drives which axis; switch back to table.
Drag a column header to reorder columns (e.g. drag a number column to position 0).
Trigger a rebuild that re-derives columns — freeze a column, hide/show a column, or add a custom column.
Check: table cells still under the correct header, and the chart still charts the correct columns. Before the fix, this is exactly where values would shift under the wrong header.
2. Medium — edit-mode reset

On a dashboard tile in view mode, sort a column (click a header to sort desc).
Enter dashboard edit mode.
Open the Network tab (or console-log formattedTableParams) and trigger a re-query in edit mode (e.g. change a filter).
Check: the outgoing request's orders should NOT carry the stale view-mode sort — before the fix it silently leaked through. This one is hard to see purely visually; devtools/network inspection is the practical way.
3. Medium — forced remount defeating the column-state guard

In view mode, apply a header filter on a tile so state differs from the saved tableFilters.
Enter edit mode (this should trigger the pending refetch + intended remount).
Once settled, freeze a column and/or drag-reorder columns.
Add or edit a custom column (triggers onColumnChange/isSelfColumnChange).
Check: the frozen/reordered column state survives the custom-column update. Before the fix, this second action could trigger an unwanted remount that wiped it.
4. Low — dead CSS

No user-visible behavior either way (the classes were never emitted). Nothing to check in the UI; confirmable only by grep, which I already did.

src/components/QueryOutput/QueryOutput.js
```diff
@@ -4071,7 +4168,7 @@ export class QueryOutput extends React.Component {
           authentication={this.props.authentication}
           autoQLConfig={this.props.autoQLConfig}
           dataFormatting={this.props.dataFormatting}
-          columns={this.state.columns}
+          columns={this.getOrderedTableColumns()}
```
alexzhou-chata, line 4171: **Medium — reordered `columns` prop breaks view-mode client-side sort/filter *types***

The display-only reorder is right for the chart/pivot/drilldown paths, but the comment's premise — *"ChataTable's Tabulator instance keys everything off `col.field` rather than array position"* — doesn't hold for ChataTable's **client-side** filter/sort path.

`getOrderedTableColumns()` (the reordered array) is passed here as `columns`. In view mode with a non-pivot, non-infinite-scroll (limited-data) table, `ChataTable.queryFn` routes through `clientSortAndFilterData` (and `getRows` → line ~979), which does:

```js
const filterColumnIndex = this.props.columns.find((col) => col.id === filter.id)?.index   // data-position index
filterDataByColumn(data, this.props.columns, filterColumnIndex, ...)
// ...and for sort:
const sortColumnIndex = this.props.columns.find((col) => col.id === ...)?.index
sortDataByColumn(data, this.props.columns, sortColumnIndex, ...)
```

Both helpers read the cell as `row[columnIndex]` (correct — data-ordered) **but derive the column *type* via `columns[columnIndex]`** (array-position lookup). Once `columnOrder` is set — a saved tile with persisted order, or a live drag — array position ≠ `col.index`, so `columns[columnIndex]` returns the *wrong* column:

- **Sort:** `sortDataByColumn` picks `isColumnNumberType`/`isColumnDateType` from the wrong column → a numeric column sorts lexicographically, a string column runs through numeric/NaN sort.
- **Filter:** `filterDataByColumn` uses the wrong `colType` → wrong coercion (e.g. DATE range logic applied to a QUANTITY column).

Cell values are still correct, so it's wrong *ordering/filtering*, not wrong data. Edit mode is safe (`queryFn` forces server-side, ChataTable.js:908) and ajax/infinite-scroll is safe — it only hits the view-mode client-side tables this feature reorders.

**Fix (in `clientSortAndFilterData`):** pass a data-ordered column list to the two helpers, e.g. `const cols = [...this.props.columns].sort((a, b) => a.index - b.index)` and use `cols` in the `filterDataByColumn`/`sortDataByColumn` calls; or resolve the column inside those calls by `id`/`field` instead of array position. A regression test with a reordered `columnOrder` + a numeric-vs-string column sort in view mode would lock this down.

alexzhou-chata (commented):
## Re-review after the `AQLP-642 - pr fixes` commits

Thanks for the updates — re-checked all three fix commits against my earlier review.

- ✅ **[High] Column reorder breaking the field↔row-data invariant** — fixed as recommended: `reorderColumnsByUserOrder` is now display-only (via `getOrderedTableColumns()` for the Tabulator `columns` prop), `state.columns` / `queryResponse.data.data.columns` stay in canonical data order, and `onColumnFreezeChange` no longer permutes the array. Bonus: column order + frozen columns are now persisted to and seeded from the tile config. **One residual edge remains — inline comment below.**
- ✅ **[Medium] Edit-mode `formattedTableParams` reset never ran for tiles** — fixed: keyed on `isEditing || isDashboardEditing`, and you also reset `this.tableParams` so a stale view-mode sort/filter can't leak back via `onTableParamsChange`. Nice — that closes the deeper part of the finding too.
- ✅ **[Medium] `_forceRemountOnNextResponse`** — fixed exactly as suggested: removed from the `isTopExecuting`-gated `componentDidUpdate` branch and applied in `setTopExecuted` once the refetch settles.
- ✅ **[Low] Dead split-pane CSS** — removed.

One **new medium** edge introduced by the reorder fix is inline. Minor: the split-view *JS* scaffolding (`getPaneBannerType`, the `tile?.splitView` guards in `DashboardTile.js`) is now fully dead since its CSS was removed — worth dropping too, or wiring up, whenever split-view lands.

<sub>🤖 Automated re-review with [Claude Code](https://claude.com/claude-code). Please sanity-check before acting.</sub>

alexzhou-chata (commented):
## Re-review of `6ea064c4` — resolved ✅

The view-mode client-side sort/filter finding is fixed correctly. `clientSortAndFilterData` now builds `dataOrderedColumns = [...this.props.columns].sort((a, b) => a.index - b.index)` and hands *that* to `filterDataByColumn`/`sortDataByColumn`, while still resolving `columnIndex` from `col.index`. Since those helpers do `columns[columnIndex]` for the column type, sorting the display array back into `.index` order makes the metadata lookup line up with the data again — equivalent to the pre-reorder invariant. The pivot branch is correctly left untouched (its columns aren't reordered and it keys off `field`). The two regression tests (numeric sort not lexicographic, numeric-equality filter) exercise exactly the failure mode and are meaningful.

That clears every finding from all three review rounds — nothing further from me on the areas I flagged. (Note: I verified by inspection; I didn't run the suite in this environment.)

<sub>🤖 Automated re-review with [Claude Code](https://claude.com/claude-code).</sub>

alexzhou-chata (approved):

PR #1393, opened 2026-07-20
Billing hooks

Shiv-Chata:


alexzhou-chata:
Reviewed the full diff (41 files, +1946/−62). Overall this is a well-built feature: the fail-open gating design is right, the error classification is verified against what the fetch helpers actually reject with, and test coverage is strong. A handful of things worth addressing before merge, ordered by impact.

## Overview

- Four public billing data hooks (`useBillingCustomerKey`, `useBillingUsage`, `useBillingHistory`, `useBillingQuotaUpdate`) plus formatting/storage helpers, exported from the package root with README + `index.d.ts` coverage.
- A MagicWand billing gate: proactive block when `quota_status === 'at_or_over_quota'`, reactive classification of `BILLING_USAGE_CEILING_REACHED` / `BILLING_USAGE_UNAVAILABLE` errors, threaded through DataMessenger, Dashboard tiles, drilldowns, and notifications behind a new `enableBillingGate` prop with an `onQuotaExceeded` callback.
- Tooling: Node 16 → 20, `--experimental-global-webcrypto` dropped from the build script, `.npmignore`, version bump.

## What looks good

- **Fail-open design is correct and tested.** The gate only blocks on a definite `at_or_over_quota`; unknown/loading/error states never block, and `BILLING_USAGE_UNAVAILABLE` never claims over-quota. The tests lock all of this in, including the gate-off fallback to existing generic error handling.
- **The error classifier's premise checks out.** `fetchLLMSummary`/`fetchLLMSummaryQuote` in autoql-fe-utils do reject with the unwrapped `error.response.data`, so reading `error?.data?.code` + `outcome` in `getMagicWandBillingErrorState` is right, and requiring the code+outcome pair before claiming over-quota is appropriately conservative.
- **The `ChatMessage` default-export change (class → forwardRef wrapper) is safe as used.** `ChatContent` reaches into the ref (`._isMounted`, `.setState` at `ChatContent.js:476`) and still gets the class instance through the wrapper; `animatedMessageIds` isn't touched outside the file; tests were updated to the named import.
- Hooks handle unmount races (`isActive` guards), storage helpers tolerate unavailable `localStorage`, and the `index.d.ts` switch to importing `Authentication` from autoql-fe-utils resolves correctly (it's exported there).

## Issues worth addressing

1. **Unbounded duplicate requests — no caching or dedup across hook instances.** `useMagicWandBillingGate` runs in the default-export wrapper of *every* `ChatMessage` and every gated `OptionsToolbar`, and each instance independently fetches customer-key + current-period usage on mount. With the gate enabled, a DM session with 30 message bubbles fires ~60 requests (text/intro bubbles included); a 20-tile dashboard fires 40. Suggest either (a) a module-level in-flight/promise cache keyed by `domain + apiKey` with a short TTL, or (b) hoisting the hook to one place per surface (ChatContent / Dashboard) and threading `quotaStatus`/`billingExecutionType` down — the inner class components already accept them as props, so most of the plumbing exists.

2. **`quotaStatus` is fetched once per mount and never refreshed.** Going over quota mid-session is covered by the reactive path, but the reverse isn't: after a user raises their quota (via `onQuotaExceeded` or another tab), mounted components keep proactively blocking until remount (e.g. DM close/reopen). `useBillingUsage` already supports `refreshKey`, but the gate never uses it. Consider refetching when the summary popover/modal opens, or after the increase-quota flow completes.

3. **Root exports missing from `index.d.ts`.** `src/index.js` re-exports everything from `hooks/billing`, which includes `useMagicWandBillingGate`, `getMagicWandBillingErrorState`, and `MAGIC_WAND_BILLING_GATE_MESSAGES`, but none of these are declared in `index.d.ts` — TS consumers importing them will get "has no exported member" errors. Either add declarations or stop exporting them from the root if they're meant to be internal.

4. **`authentication: null` crashes the public hooks.** The `= {}` defaults in `useBillingCustomerKey`/`useBillingUsage`/`useBillingHistory` only apply to `undefined`; an explicit `null` reaches the dependency array (`authentication.apiKey`) and throws during render. Since these are documented public hooks, a null from a consumer during auth bootstrapping is plausible. Cheap fix: normalize (`authentication ?? {}`) or use optional chaining in the deps.

5. **Over-quota copy is duplicated.** `MagicWandBillingGateNotice` hardcodes "Your most recent Auto Analyze usage put this account at or over its monthly quota." while `MAGIC_WAND_BILLING_GATE_MESSAGES.over_quota` carries the same sentence — the DM bubble and the popover notice can drift apart on the next wording change. The `unavailable` branch already uses the constant; the over-quota branch should too (split the constant if the notice needs the shorter form).

6. **Currency formatting hardcodes USD/en-US** in `formatMicrosAsCurrency`/`currencyInputFromMicros`, while `BillingQuotaUpdateResponse` includes a `currency` field. Worth accepting an optional currency/locale parameter now (defaulting to USD) so the signature doesn't have to break later.

## Minor notes

- The proactive-check + reactive-classification block is now repeated across ~6 call sites (ChatMessage ×3, OptionsToolbar, SummaryModal). A small shared helper would keep the gating rule in one place.
- `MagicWandBillingGateNotice` appears in response to a user action but has no `role='alert'`/`aria-live`, so screen readers won't announce it.
- Each blocked Auto Analyze click in the DM appends another identical gate message bubble — consider not re-adding if the last message is already the gate notice.
- The exact-string match on the "Billing is required for this request…" message is acknowledged as fragile in the code comment — worth a follow-up ticket asking Query Controller to return a discriminating code so it can be removed.
- The billing endpoints use raw `fetch` with hand-rolled URL builders while every other API call in the repo goes through autoql-fe-utils (axios). If that's deliberate to keep this package self-contained, fine — otherwise consider moving them into autoql-fe-utils later for consistency (interceptors, cancellation).
- `.npmignore` is effectively redundant given the `"files"` allowlist in package.json.
- Node 16 → 20 + dropping the webcrypto flag: the build now requires Node ≥ 19 (where global webcrypto is unflagged) — make sure CI/publish environments move in lockstep since there's no workflow file in-repo to update.
- `updateQuota` throws generic `Error('Failed to update billing quota')` without the status or response body attached, so consumers can't distinguish auth failures from validation errors when building UI on top of it.

alexzhou-chata:
Re-reviewed the update (58b41b3b). The substantive points from the first pass are all addressed, and the mechanisms hold up:

- **Request dedup** — the in-flight collapse in `dedupeBillingRequest` is the right shape: keyed by domain/apiKey/token (+ customer key/refreshKey for usage), entries removed on settle so it never serves stale data, rejections propagate to all callers, and each hook's `isActive` guard still protects unmounted instances. Nice touch that a refresh broadcast also coalesces — every gate bumps to the same `refreshKey`, so the refetch is a single request rather than a stampede.
- **Staleness** — the refresh bus plus the automatic `refreshBillingUsage()` after a successful `updateQuota` closes the "raised quota but still blocked until remount" gap, and exporting `refreshBillingUsage` gives hosts an escape hatch for out-of-band changes. The lazy `useState(getBillingUsageRefreshVersion)` + subscribe-on-mount pattern is correct for this React version.
- **`index.d.ts`** now declares the gate hook, classifier, messages (including `over_quota_summary`), and `refreshBillingUsage` — with a useful doc comment on the latter.
- **Null-auth normalization**, the **shared over-quota copy**, and the **currency/locale parameters** all look good.

Three follow-ups, one of which is a real bug in the new code:

1. **The fraction-digits probe in `currencyInputFromMicros` is a no-op.** `getCurrencyFormatter` always constructs with `maximumFractionDigits: 2`, and an explicitly provided max wins in `resolvedOptions()` — so the probe returns 2 for every currency (JPY resolves 2 instead of 0, BHD resolves 2 instead of 3). Net effect: inputs still always render two decimals, and `formatMicrosAsCurrency` truncates 3-decimal currencies. Dropping the hardcoded `maximumFractionDigits: 2` from the formatter options fixes both — currency defaults already produce 2 for USD.

2. **`useBillingQuotaUpdate` didn't get the null-normalization** the three read hooks received — it still uses a destructuring default, so `authentication: null` makes `updateQuota()` throw a `TypeError` from `hasBillingAuthentication(null)` instead of the intended `Error('Missing billing authentication')`. Same one-line `options.authentication ?? {}` treatment would make it consistent.

3. **The new behavior is untested.** The dedup and the refresh bus are the core of this commit, and both are easy to regress silently (e.g. someone later "simplifies" the cache key or the settle-cleanup). Given the bar the rest of the PR set, a few targeted tests would be worth it: two harnesses with identical auth mounting concurrently → one fetch; a rejected deduped request settling all callers to their error states; `refreshBillingUsage()` clearing a mounted gate's `at_or_over_quota` after a quota update.

The remaining minor notes from the first pass (`role='alert'` on the notice, repeated gate bubbles in the DM) weren't part of this update — fine to defer, just noting they're still open.

alexzhou-chata:
Checked 60b31439 — both fixes look right:

- **Fraction digits**: with the hardcoded `maximumFractionDigits: 2` removed, the currency defaults drive the output correctly — USD behavior is byte-for-byte unchanged (`$12.50` / `12.50`, so the existing formatting tests still pass), JPY now resolves 0 decimals (`¥13` / `13`), and 3-decimal currencies like BHD get their full precision (`BHD 12.346` / `12.346`).
- **`useBillingQuotaUpdate`** now normalizes `authentication` the same way as the read hooks, so a `null` produces the intended `Error('Missing billing authentication')` instead of a `TypeError`.

The one remaining item from the last pass is test coverage for the dedup and refresh-bus behavior (concurrent mounts → one fetch; rejected deduped request settles all callers; `refreshBillingUsage()` clears a stale over-quota gate). Nothing else outstanding from my side beyond the two deferred minors (`role='alert'` on the notice, repeated gate bubbles in the DM). LGTM once the tests are in.

alexzhou-chata:
Checked a798e1bb — this covers everything I asked for, and then some:

- `dedupeBillingRequest` unit tests hit the exact regression risks called out earlier: concurrent callers collapse to one factory invocation, rejections propagate to every sharer, and the cache clears on settle for both the resolve and reject paths (so retries re-fetch). The distinct-keys case is a nice addition.
- The integration tests exercise the real hook path: three concurrent `useBillingCustomerKey` mounts → one fetch, and a rejected shared request lands every instance in its error state.
- The gate tests cover the two behaviors that matter most for the feature: a DM-full-of-bubbles scenario (three gates → one customer-key fetch + one usage fetch) and `refreshBillingUsage()` clearing a stale `at_or_over_quota` on a still-mounted gate — correctly asserting the customer-key fetch does *not* re-fire, only usage.

I also ran the suites locally against this head: all four billing hook suites pass (33 tests), and the component/util suites (`MagicWandBillingGateNotice`, `SummaryModal`, `ChatMessage`, `OptionsToolbar`, `billingFormatting`, `billingCustomerKey`) are green as well.

Nothing outstanding from my side — LGTM. (The two cosmetic items — `role='alert'` on the notice and repeated gate bubbles in the DM — are fine as follow-ups whenever convenient.)

alexzhou-chata (approved):

PR #1394, opened 2026-07-20
Develop

eleon-chata-ai:
develop to master


alexzhou-chata:
## Code review — PR #1394 (develop → master, `v8.31.0`)

Reviewed the source changes (excluding the large Prettier reformatting churn and lock-file). Findings below were verified against the actual `develop` code. Overall this is solid, well-commented work — the freeze/reorder-persistence and locked-filter reworks are careful. A few items are worth addressing before this ships.

### What this PR does
- **Column freeze + drag-reorder persistence** across tile remounts/rebuilds (matched by column *name*, not index) — `QueryOutput`, `ChataTable`, `TableWrapper`, `DashboardTile`.
- **Locked/base filters** rework: replaces the DOM-input-disabling approach with a `lockedFilters` prop threaded through `QueryOutput` → `ChataTable`; edit⇄view mode filter/sort resync.
- **DashboardTile lifecycle**: immediate `queryId` flush, self-column-change remount suppression, split-view scaffolding, forced-remount-on-refetch.
- **Dashboard**: extracted pure `isResponseFailed`/`isTileFailed`/`isTileDirty` helpers, dashboard-identity-change state reset, `queryId` propagation.
- **MEDIAN** custom column (agg + window), **ResizeObserver** error suppression + `colorScale` in `ChataChart`, drilldown filters now include groupBy values, data-alert label scoping, and a large batch of new tests.

---

### 🟠 Medium — worth fixing before merge

**1. Cross-dashboard tile leak on undo** — [`Dashboard.js:227-240`](src/components/Dashboard/Dashboard.js#L227)
The new `dashboardId`-change reset block clears `pendingResetTiles`, `baselineQueryIds`, `tileLog`, etc., but **not** `this.pendingResetUndoTiles` / `this.pendingResetHistory`. `canUndo()` short-circuits on `if (this.pendingResetUndoTiles) return true` (L953) and `undo()` restores from it (L901) via `flushOnChange(restoredTiles)`.
*Scenario:* reset a tile in edit mode on dashboard A (sets `pendingResetUndoTiles` = A's snapshot at L1139), host swaps `dashboardId` to B while still editing, then undo → **dashboard A's tiles get published into dashboard B** — exactly the leak this block's comment says it prevents.
*Fix:* add `this.pendingResetUndoTiles = null; this.pendingResetHistory = null` (and `clearTimeout(this.discardResetChangesTimer)` / `resetGuardTimer`) to the reset block, mirroring L348-350.

**2. Split-view tiles suppress dirty/failed warnings with no replacement** — [`DashboardTile.js:426`](src/components/Dashboard/DashboardTile/DashboardTile.js#L426), L1661, L1688
For `tile.splitView`, `shouldShowDirtyBadge()` returns `false` and the `failed` class + failed-badge are suppressed ("split view uses per-pane banners"). But the replacement `getPaneBannerType()` (L435) is **never called in the render path** — `renderContent`/`renderTopResponse` only render the top response, and `secondResponseRef` is never assigned, so `wasBottomFilteringBeforeRemount` (L246) is dead too.
*Effect:* if a split-view tile is dirty/failed in edit mode, it shows **no indicator at all**, so a user can save a stale/unexecuted split-view tile with no warning. This looks like a half-landed feature. Please confirm whether split-view tiles render through this component in production — if the second pane/per-pane banners aren't shipping in this PR, keep the badge fallback for split-view tiles; otherwise wire `getPaneBannerType` (+ `wasBottomFilteringBeforeRemount`) into the render.

---

### 🟡 Low

**3. `clearSort()` runs unguarded next to the try/caught `clearHeaderFilter()`** — [`ChataTable.js:404`](src/components/ChataTable/ChataTable.js#L404)
`clearHeaderFilter()` is wrapped in try/catch precisely because a stale column ref can make Tabulator throw synchronously. The immediately-following `clearSort()` touches the same bookkeeping but isn't wrapped — if it throws, `setFilters(baseFilters)`/`setSorters(baseSort)` below never run, leaving the table inconsistent on the view⇄edit toggle. Wrap it in the same guard.

**4. `addTile` can overwrite a valid `queryId` with `undefined`** — [`Dashboard.js:842`](src/components/Dashboard/Dashboard.js#L842)
`const queryId = content.queryResponse?.data?.data?.query_id` then `tile = { ...tile, ...content, queryId }`. In the else-branch, `content` may carry a valid `queryId` but be spread-over with `undefined` when the response has no `data.data.query_id` (e.g. an error/reconstructed response), weakening dirty detection. Guard it like L1201: `...(queryId !== undefined ? { queryId } : {})`.

**5. `_forceRemountOnNextResponse` can stay armed after a cancelled refetch** — [`DashboardTile.js:273`](src/components/Dashboard/DashboardTile/DashboardTile.js#L273)
Set true when entering edit with `viewModeFiltersDiffer`, cleared only in `setTopExecuted()`. On the `REQUEST_CANCELLED_ERROR` path `setTopExecuted` never runs, so the flag persists and a later unrelated top-query execution bumps `queryResponseVersion` → an unnecessary full `QueryOutput` remount. Benign (no data loss) but easy to clear in the cancel/catch path.

**6. Data-alert label fetch default flips to project-scoped** — [`DataAlertModal.js:140`](src/components/Notifications/DataAlertModal/DataAlertModal.js#L140), `CustomFilteredAlertModal.js:121`
`getLabels()` now calls `getAllDataAlertsLabelsByProject` whenever `isManagementPortal` is falsy, and it defaults to `false`. No component in the repo passes `isManagementPortal={true}`, so **every existing consumer** switches from `getAllDataAlertsLabels` to the by-project endpoint. (`getAllDataAlertsLabelsByProject` is exported by `autoql-fe-utils@1.11.27`, so no crash.) If management-portal consumers relied on the all-labels list, please confirm this default flip is intended and/or note it in the changelog.

---

### 🔵 Please confirm intent

**7. Locked/base filter inputs are no longer disabled in view mode** — `ChataTable.js` / `ChataTable.scss`
`disableBaseFilterInputs()`/`enableBaseFilterInputs()` and `.react-autoql-base-filter-disabled` were removed. `baseFilters` now comes from `props.lockedFilters` but nothing re-disables those header inputs, so a viewer could edit/clear a locked filter and re-run with the lock altered. Is immutability enforced elsewhere now, or is this an intended UX change?

**8. Freezing a column now fully remounts the table** — [`QueryOutput.js:4063`](src/components/QueryOutput/QueryOutput.js#L4054) → `onColumnFreezeChange` bumps `columnChangeCount`, which regenerates `this.tableID` (uuid) used as `ChataTable`'s `key`. That tears down and rebuilds the whole Tabulator (re-runs the data pipeline / server refetch for ajax tables, resets scroll) on every freeze toggle — despite the `onFreezeColumnClick` comment saying "Persist via parent + `setColumns()` rebuild" (the in-place `setColumns` path never runs). Likely a deliberate tradeoff, but the comment is misleading and the flicker/refetch cost on large tables is worth a look.

**9. `min-width:300px` variables/operators panel is now conditionally rendered** — [`CustomColumnModal.js:2191`](src/components/AddColumnBtn/CustomColumnModal.js#L2191) — the previously always-present 300px flex spacer no longer renders when the function-config panel is open. Almost certainly a beneficial cleanup; just worth a visual check that the config-panel width still looks right.

---

### 🧪 Test coverage
- [`QueryOutput-onChartClick.test.js:252`](src/components/QueryOutput/__tests__/QueryOutput-onChartClick.test.js#L252) & `:308` — the two drilldown `groupByFilters` tests are **tautological**: they build `groupByFilters` inline, call `onDrilldownEnd({ drilldownFilters: groupByFilters })` (or assert on the locally-built object) and never drive `processDrilldown`'s API branch where the real `groupBys.map(...)` mapping lives. A regression in the production mapping would not fail these. Drive the real path instead.
- [`Dashboard.resetTile.test.js:85`](src/components/Dashboard/__tests__/Dashboard.resetTile.test.js#L85) — several `describe` blocks named for specific behaviors (`addTileStateToLog during reset`, `handleDiscardEvent`, `componentDidUpdate isResettingTile guard`, `resetTileStateLog`) only `expect(wrapper).toBeDefined()`. The named logic could break entirely and these pass. (The `pendingResetTiles`/undo/redo tests in the same file are genuine — nice.)
- [`ChataTable.test.js:504`](src/components/ChataTable/ChataTable.test.js#L504) & `:672` — real wall-clock `setTimeout(…, 10)` / `100` to await internal timers; likely flaky on loaded CI. Prefer `jest.useFakeTimers()` + `advanceTimersByTime`.

### ⚪ Nits
- `hasData` is unused (only referenced in a commented-out block) — [`OptionsToolbar.js:1035`](src/components/OptionsToolbar/OptionsToolbar.js#L1035).
- `_columnChangeQueryId` is never reset — [`DashboardTile.js:1047`](src/components/Dashboard/DashboardTile/DashboardTile.js#L1047); resetting it after it's matched in `componentDidUpdate` would make the one-shot intent explicit.
- `ChataChart` registers one `window` `'error'` listener per instance (add/remove is symmetric, so no leak) and the handler's `return false` is a no-op — a single module-level ref-counted listener would be cleaner.

### General
- The diff mixes large Prettier-only reformatting with real logic changes in the same files (`CustomColumnModal`, `QueryOutput`, `ChataTable`), which makes review harder. Consider splitting formatting-only churn into its own commit/PR next time.
- Nice touch: the circuit-breaker comments, the `name`-based (not index-based) matching for freeze/reorder, and the pure `isTile*` helper extraction with dedicated tests.

*Automated review — please sanity-check the medium items (#1, #2) and the intent questions before merge.*

alexzhou-chata (approved):

PR #1395, opened 2026-07-27
AQLP 829 

eleon-chata-ai:
- Support multiple projects in single dashboard
- AQLP-862 - None of these response remains when I try to delete it
- bug fixes


alexzhou-chata:
## PR Review

### Overview
This PR bundles three concerns: (1) **AQLP-829** — per-tile project support for multi-project dashboards (new `projectSelectList` / `getAuthenticationForProject` / `showProjectIndicator` props, tile-level `projectId` with badge, edit-mode picker modal, per-project auth resolution and an auth-wait guard before firing queries), (2) **AQLP-862** — "None of these" messages now carry `queryMessageID` so deleting the suggestion response cascades to the whole thread, and (3) perf/bug fixes — rAF-throttled drawer resizing with proper cleanup, `ChatContent.shouldComponentUpdate` render skip during resize, `columnOverrides` preserved in `QueryOutput` dataConfig, and `ConfirmPopover` no longer portals into `popoverParentElement`.

### What looks good ✅
- **Consistent auth threading**: `getTileAuthentication()` / `getTileAutoQLConfig()` are applied everywhere it matters — `processQuery`, `fetchAutocomplete`, `QueryOutput`, `OptionsToolbar`, `ReverseTranslation`, `FollowOnModal`. Since dashboard drilldowns run inside `QueryOutput`, drilldowns inherit the tile-scoped auth too.
- **The auth guard is wired correctly end-to-end**: the fabricated `1.1.401` response flows through `endTopQuery` → `hasError()` (non-2xx reference id → error render), and the spinner set in `processTile` is cleared via `setTopExecuted`, so a tile that never gets a token lands in a clean error state instead of a doomed 403 request.
- **Test coverage is strong**: 40 new tests in `DashboardTile.multiProject.test.js` plus additions to the Dashboard/DashboardTile/DataMessenger suites. I ran all four touched suites locally at this PR's head — **336/336 pass**.
- The resize work is careful: rAF throttling with captured coordinates, the `e.buttons === 0` guard for mouseup-outside-window, rAF cancellation + listener removal on unmount, and the `ChatContent` render skip is scoped tightly (`isResizing` on both prop snapshots **and** `nextState === this.state`), so messages arriving mid-drag still render.
- The `columnOverrides` fix in `QueryOutput.js` closes a real silent-data-loss hole on save.

### Issues

**1. Leaked local-dev artifact in `package-lock.json` (should fix before merge)**
The lockfile now contains an `"../autoql-fe-utils"` entry marked `"extraneous": true` — that's your local sibling checkout leaking into the committed lockfile. Please regenerate the lockfile without the local `file:` install present (the diff should only show the version bump).

**2. `projectId` string/number comparisons are inconsistent (medium)**
`tile.projectId` and `projectSelectList[].projectId` are both typed `string | number`, but comparisons are split:
- `renderProjectBadge` / `hasNonDefaultProject` string-coerce (`` `${a}` === `${b}` ``) ✅
- `getSelectedProjectName`, the modal `Select` value match (Select uses strict `===` internally), `confirmDisabled`, and the `componentDidUpdate` re-run check all use strict equality ⚠️

If a saved tile has a numeric `projectId` while `projectSelectList` uses strings (or vice versa — PropTypes explicitly allow both), the modal won't preselect the current project, `confirmDisabled` won't block a no-op "change" to the same project, and a pure type flip (`2` → `'2'`) re-runs the query and dirties the dashboard. Recommend normalizing to string in one place (e.g. compare via a small `isSameProjectId` helper) and using it everywhere.

**3. Unmount race in `waitForTileAuthentication` (minor)**
On unmount mid-wait the promise resolves `true`, so `processTile` proceeds into `processTileTop`, which immediately calls `this.setState({ isTopExecuting: true, queryResponse: null })` with no `_isMounted` guard and then fires a network request on an unmounted tile. Suggest resolving a distinct sentinel (or re-checking `this._isMounted` in the `.then`) and bailing out instead.

**4. `onSuggestionClick` bypasses the auth wait (minor)**
Validation-suggestion button clicks call `processTileTop` directly, skipping `waitForTileAuthentication`. It still uses `getTileAuthentication()` (falling back to dashboard auth if the token isn't cached yet), and in practice a token usually exists by then, but routing it through `processTile` would keep the guarantee uniform.

**5. `ConfirmPopover` portal change — please confirm intent (question)**
Dropping `parentElement={this.props.popoverParentElement}` (while keeping `boundaryElement`) moves the popover portal to the default container for **all** consumers — DataMessenger's clear-messages button, NotificationFeed's delete confirmations, and `Chip`. If this fixes a clipping bug, 👍 — but has it been visually checked inside the drawer and in fullscreen mode where stacking contexts/z-index can differ? A one-line note on which bug this fixes would help future readers.

**6. Nits**
- `src/index.d.ts`: `projectSelectList` types `projectId` as `string` only, while PropTypes accept `string | number` — make them agree.
- `ChatMessage` uses the new `queryMessageID` prop but doesn't declare it in `propTypes`.
- `stopResizingDrawer` is now called from `componentWillUnmount` after `_isMounted = false`; if unmounted mid-drag it will `setState` on an unmounting component (dev warning). An `_isMounted` check before the `setState` would silence it.
- UX note: a tile whose token never arrives spins for the full 15s before showing the auth error. Fine as a bound, but if the host app knows a token fetch failed, a promise-returning `getAuthenticationForProject` would let you fail fast (future consideration, not a blocker).

### Verdict
Solid, well-tested feature work — the multi-project plumbing is thorough and the guard rails (auth wait, error fallback, re-run on project change) hang together. #1 should be fixed before merge, #2 is worth addressing now while the comparison sites are fresh, and #5 just needs confirmation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alexzhou-chata:
## Follow-up Review (6508ff7)

Re-reviewed the update — all four actionable items from the first pass are resolved, and I verified each one at the new head (**348/348 tests pass** across the four touched suites, including the 12 new ones):

- ✅ **Lockfile**: the extraneous `"../autoql-fe-utils"` entry is gone; only the registry-resolved `autoql-fe-utils@1.11.27` remains.
- ✅ **`projectId` comparisons**: the nullish-aware `projectIdsEqual` helper is now used consistently (re-run check, `topRequestData` sync, badge, selected name, `hasNonDefaultProject`, `confirmDisabled`), and `resolveListProjectId` fixes the modal `Select` preselection for type mismatches. The new tests cover the type-flip cases directly (no re-run on `2` → `'2'`, no false structural change, Modal stays disabled on a same-project type flip). Nice.
- ✅ **Unmount race**: `runWithTileAuthGuard` bails after the wait when unmounted — no query, no fabricated error state. Tested.
- ✅ **`onSuggestionClick`**: now routed through the same auth guard with the loading state set. Tested both before/after token arrival.
- ✅ Nits: `index.d.ts` now `string | number`, `queryMessageID` declared in `ChatMessage` propTypes, `_isMounted` guard added in `stopResizingDrawer`.

Two small things remain:

1. **ConfirmPopover portal change** (from the first review, still open): dropping `parentElement` affects all consumers (DataMessenger clear-messages, NotificationFeed delete, Chip). Just confirming it was visually checked inside the drawer/fullscreen stacking contexts — a one-line note on which bug it fixes would be enough.
2. **Tiny nit, non-blocking**: in `componentDidUpdate`, the projectId-change branch calls `this.processTile({ query: this.props.tile.query })` bare, while the sibling edit-mode refetch a few lines below uses `this.processTile()?.catch?.(() => {})`. A failed re-run after a project change will log an unhandled promise rejection (the tile's error state still renders correctly). Same one-liner would silence it.

LGTM once #1 is confirmed. 👍

🤖 Generated with [Claude Code](https://claude.com/claude-code)

eleon-chata-ai:
done

src/components/Dashboard/Dashboard.js
```diff
@@ -150,6 +150,19 @@ class DashboardWithoutTheme extends React.Component {
     onQuotaExceeded: PropTypes.func,
     enableFollowOnQuery: PropTypes.bool,
     enableResetQuery: PropTypes.bool,
+    // List of projects tiles can be assigned to (multi-project dashboards). When omitted, no per-tile project picker is shown.
```
NikkiMeganMoore, line 154: Can we add an additional check here - if there is only **one** project in the array, we should display the original UI (no project picker)

NikkiMeganMoore (changes requested):

src/components/QueryOutput/QueryOutput.js
```diff
@@ -125,8 +125,7 @@ export class QueryOutput extends React.Component {
     this.queryResponse = _cloneDeep(response)
     this.columnDateRanges = getColumnDateRanges(response)
     this.queryID = this.queryResponse?.data?.data?.query_id
-    this.drilldownQueryID =
-      this.queryResponse?.data?.data?.drilldown_query_id || this.queryResponse?.data?.data?.query_id
+    this.drilldownQueryID = this.queryResponse?.data?.data?.query_id
```
NikkiMeganMoore: Isn't this supposed to be using the drilldown query id? Curious why you are changing it

eleon-chata-ai: there was a change requested here because the query_id should be the same as the drilldownid now, so drill down id is not actually needed anymore

src/components/Dashboard/Dashboard.js
eleon-chata-ai: fixed

PR #1397, opened 2026-08-05
Column visibility client side

NikkiMeganMoore:


eleon-chata-ai:
[HIGH] src/components/QueryOutput/QueryOutput.js:2211-2216 — When the table sort is cleared, syncChartSortWithTableSort unconditionally calls this.onAxisSortChange('x', this.tableConfig?.stringColumnIndex, null) regardless of which axis/column was actually synced last. For heatmap/bubble charts, onAxisSortChange preserves prevState.axisSorts and only deletes the ${axis}-${columnIndex} key it's given — so if the last sync was a y-<numberColIndex> entry (a number-column table sort), clearing the table sort deletes the wrong key (x-<stringColIndex>) and leaves the stale y-… sort permanently stuck in chart state.

[HIGH] src/components/QueryOutput/QueryOutput.js:2243-2245 (interacting with 4432) — onAxisSortChange always does setState({ chartID: uuid() }), and <ChataChart key={this.state.chartID} hidden={!isChartType(...)} .../> stays mounted (just CSS-hidden) while the table view is active. So sorting the table now force-remounts the hidden chart on every sort change, wiping any chart-local-only state (e.g. scale-toggle state in ChataColumnChart/ChataBarChart) — a new side effect that didn't exist before this PR.

[MEDIUM] src/components/QueryOutput/QueryOutput.js:2228-2233 — syncChartSortWithTableSort maps any number-column table sort to a chart y-axis "value" sort via onAxisSortChange, but AxisSortPopover.js explicitly restricts heatmap/bubble charts to string-only sort options in its own UI. The sync path produces axis-sort state the chart's own controls can never reach or clear again through the UI.

[MEDIUM] src/components/Charts/chartPropHelpers.js:17-24 — The new shared BAR_GRADIENT_OPACITY_STOPS = [0.45, 0.55, 0.65] replaces HistogramColumns's original [0.9, 0.7, 0.5] (spread 0.4) with a reversed [0.65, 0.55, 0.45] (spread 0.2) — a visible flattening of that chart's gradient contrast that should be confirmed as intentional with design, not just a byproduct of unifying the constant.

[LOW] src/components/Charts/ChataSankeyDiagram/SankeyColumnSelector.js:8 — SANKEY_VALUE_COUNT_SENTINEL = -1 is exported from ChataSankeyDiagram.js but copy-pasted as a second literal in this file with a "must match" comment instead of imported (there's a real circular-import constraint between the two files, so this needs a small shared-constants module, not a trivial fix).

[LOW] src/components/QueryOutput/QueryOutput.js:2189-2195, 2204-2207 — Two multi-line comments (6 and 4 lines) justify a fragile assumption that generatePivotData and its callees stay synchronous forever; this should be a 1-line comment, and the underlying fragility itself is a latent landmine if that assumption ever breaks.

[LOW] Process — bundling unrelated chart/table/Sankey/version changes into one PR titled "column visibility" makes this hard to review atomically and risky to revert if one piece needs to come out post-merge.

eleon-chata-ai (approved):

